Issue #8, opened 2015-05-24
Code style


jomo:
I noticed there's tabs and spaces mixed all over the place. Can we agree on a style to use everywhere?
My editor (sublime text + eslint) also starts highlighting every other line because there's missing semicolons or other issues.

![screenshot](https://i.imgur.com/M1FEr3n.png)

I personally prefer 2 spaces for indentation, I think that's what most node projects seem to use as well.
I recommend that everyone uses [eslint](https://github.com/eslint). There are lots of [integrations for different editors](http://eslint.org/docs/user-guide/integrations).

rmccawl:
Another recommendation would be to use something like an .editorconfig file where you can specify how the text-editor handles specific file types based on the extensions.

You can configure tabbing, spaces, line-endings and much more.

http://editorconfig.org/

jomo:
indeed, editorconfig is really nice :+1:

hans-sperling:
Right! "editorconfig is really nice" :+1:

PR #11, opened 2015-05-24
Add eslint file to project

jomo:
I imported the nodejs eslint file, but disabled 'strict' and 'max-len'. I think they are overkill.

If we agree on using this, I would fix the style in existing files and #8 could be closed.


jomo:
Additional changes I'd like:

0. disable [key-spacing](http://eslint.org/docs/rules/key-spacing)
0. disable [eol-last](http://eslint.org/docs/rules/eol-last)
0. add "browser" [environment](http://eslint.org/docs/user-guide/configuring#specifying-environments) (for js assets)

Let me know what you think!

jomo:
My latest commit added a few more rules that I found useful and fixes nearly all errors.
I haven't fixed errors that would change the actual logic, but I have left some commit comments where I thought they were errors.

If everyone agrees on this, I'd like to get this merged in quickly to avoid merge conflicts.

volumio:
I agree on this. Steven do you agree as well?

jotak:
+1

.eslintrc
```diff
@@ -0,0 +1,35 @@
...
+  no-fallthrough: 2
+  no-multi-spaces: [2, { exceptions: { "ObjectExpression": true } }]
+  no-reserved-keys: 2
+  no-unreachable: 2
+  padded-blocks: [2, "never"]
+  quotes: [2, "single", "avoid-escape"]
+  semi: 2
+  space-after-keywords: 2
+  space-before-blocks: 2
+  space-before-function-paren: [2, "never"]
```
ning-yu: I'd prefer `space-before-function-paren: 2`

.eslintrc
```diff
@@ -0,0 +1,35 @@
+env:
+  node: true
+  browser: true
+
+rules:
+  brace-style: 2
+  comma-spacing: 1
+  eol-last: 0
+  indent: [2, 2]
```
ning-yu: I would very much prefer `indent: [2, "tab"]`

.eslintrc
```diff
@@ -0,0 +1,35 @@
+env:
+  node: true
+  browser: true
+
+rules:
+  brace-style: 2
+  comma-spacing: 1
```
ning-yu: I'd prefer `comma-spacing: 2`

.eslintrc
```diff
@@ -0,0 +1,35 @@
...
+  eol-last: 0
+  indent: [2, 2]
+  key-spacing: 0
```
ning-yu: I'd prefer `key-spacing: 2`

.eslintrc
```diff
@@ -0,0 +1,35 @@
...
+  new-parens: 2
+  no-debugger: 2
+  no-dupe-args: 2
+  no-dupe-keys: 2
+  no-duplicate-case: 2
+  no-ex-assign: 2
+  no-extra-parens: 1
+  no-fallthrough: 2
+  no-multi-spaces: [2, { exceptions: { "ObjectExpression": true } }]
```
ning-yu: Why are `ObjectExpression`'s exempt?

.eslintrc
```diff
@@ -0,0 +1,35 @@
+env:
+  node: true
+  browser: true
+
+rules:
+  brace-style: 2
+  comma-spacing: 1
+  eol-last: 0
```
ning-yu: Let's add `eqeqeq: [2, "smart"]`

.eslintrc
```diff
@@ -0,0 +1,35 @@
...
+  space-before-blocks: 2
+  space-before-function-paren: [2, "never"]
+  space-infix-ops: 2
+  space-return-throw-case: 2
+  spaced-line-comment: 2
+  strict: 0
+  valid-typeof: 2
+
+ecmaFeatures:
+  blockBindings: true
```
ning-yu: Let's add the following:
`default-case: 2`

.eslintrc
jomo: The reason I prefer spaces is that tabs have no globally defined width.
Consider this:

```js
function() {
  var one = 1,
      two = 2;
  // ...
}
```

If you do this with tabs, it looks like this:
```js
function() {
<tab>var one = 1,
<tab>    two = 2;
  // ...
}
```

Using tabs *and* spaces in the same project is absolutely ridiculous.  
Any decent editor will allow you to insert *n* spaces when you press tab.  
2 spaces is also what's used pretty much everywhere in the node community.

----

What do the others think about this?

.eslintrc
jomo: :+1:

.eslintrc
jomo:
```js
return self.sendMpdCommandArray([
  {command: 'clear', parameters: []},
  {command: 'add',   parameters: [arrayTrackUris.shift()]},
  {command: 'play',  parameters: []}
])
```
The middle and last line would be an error otherwise

.eslintrc
ning-yu: I don't use tabs to give a defined width. I use tabs to indicate an indent, and spaces to indicate separation between objects. I always assume that anyone can have the tab size set to any value they choose, and feel you would see no difference if you set your tabs to display as two spaces wide.

.eslintrc
jomo:
```js
function name(args) {}
// instead of
function name (args){}
```
is used by node and also [suggested](http://javascript.crockford.com/code.html) by Crockford (js inventor):
> There should be **no space** between the name of a function and the  `(`
> There should be **one space** between the `)` and the `{`

.eslintrc
ning-yu: Ok that's fine. I usually don't line them up like you do.

.eslintrc
ning-yu: How about spaces between both? :)

.eslintrc
jomo: that is actually [your code](https://github.com/volumio/Volumio2/commit/2414e471e3bb024d7f822727955b5afc5a051415#diff-3c6db640873e19d10724061e8ba731e8R57) :stuck_out_tongue_winking_eye:

ning-yu: Yeah cut me some slack, I'm only human! :man:

volumio:
Ok, so it seems that you negotiated it pretty well. Since I'm not really into discipline, I will stick to what you guys came up with. 
However we now have merge conflict. If you could fix it, I'll then merge

ning-yu:
@volumio
Let's see if @jomo is okay with sticking with tabs. I don't want to proceed without everyone okay with it.

@jomo
Once we agree on the conventions, I propose that we only include the .eslintrc file in this commit, instead of linting every file.

We'll then give the owners of each file one month to make their code conform to the convention, so they can do this in parallel with any changes in work. After one month, we'll run every file through ESLint to catch anything remaining.

volumio:
I vote for tabs as well. We can decide to configure our IDEs to have a consistent spacing for tabs. 

Also agree on the one month deadline

ning-yu:
After some reflection, I think the reason I personally prefer tabs over spaces has to do with how I fundamentally think of formatting source code.

When I write code, I see it as being freely structurable in 1 dimension only. I insert extra linebreaks in order to organize code in the vertical direction, but I do not add spaces in order to align code in the horizontal direction. The indent at the beginning of each line exists only to indicate the nesting depth of the line, and is fixed by the content of the code itself, and thus is not adjustable. In this case, it makes sense to use tabs as the indent character. The boundaries here are soft, however. I will often perform horizontal arrangement of text in block comments to make them more readable.

Using spaces to indent seems to me to be more consistent with seeing code is freely structurable in 2 dimensions. In this case, one can line up objects in each line with their corresponding objects on surrounding lines to increase readability. The indent at the beginning of the line indicates nesting depth, but extra spaces can also be added to the beginning of line to provide more information on the context of the line. In this case, it does not make sense to mix both tabs and spaces as indent characters, and spaces will definitely win out.

2D structured code is absolutely beautiful, and a joy to read. On the other hand, it is more difficult to maintain, and there are more formatting variables to deal with in order to maintain consistency. 1D structured code is more difficult to read, but being minimally formatted, is also easier to write and maintain.

jomo:
my latest commit has the changes you asked for.

----

so you'd agree to use spaces?
if @volumio would also agree on spaces, then we only need to discuss the style of `space-before-function-paren`.

looking at these:
```js
function name(args) {
  // foo
}
var fun = function() {
  // foo
}

////////////////////////////////////////////////////////////////

function name (args){
  // foo
}
var fun = function () {
  // foo
}

////////////////////////////////////////////////////////////////

function name (args) {
  // foo
}
var fun = function (args) {
  // foo
}
```
I think the first is the most readable. As I said above, it's also used pretty much everywhere else (and endorsed by Crockford).
It's still quite possible to differentiate them from function calls because they have no `function` keyword:
```js
name();
```

----

Why would you want to wait a month? IMO it's much simpler to just apply all the rules at once, and then people can continue their work.
I've already done that for the whole project in this PR, btw.

volumio:
Merging, so we can proceed without conflicts.

ning-yu:
> so you'd agree to use spaces?

I don't believe we've come to an agreement yet. I prefer tabs still, until you can prove that the benefits of using space indents outweigh the additional upkeep.

I'll defer to your preference on `space-before-function-paren`.

volumio:
I'm for tab as well. Makes the code easier to read, since this is a community driven project I think we should prioritize readability over polished code.

It has been a great discussion, allowing us to have some insights on code style, but guess now we can move on and have some fun with the code !

jomo:
I don't see any "additional upkeep costs" at all, tbh.
Any decent editor allows you to insert any amount of spaces when you press tab, and to remove the same amount when you press backspace.

This setting could be automatically applied for the project with editorconfig, as suggested by @rmccawl.

-----

The true answer is, however, there is no true answer. It's a [never](http://www.jwz.org/doc/tabs-vs-spaces.html) [ending](http://blog.codinghorror.com/death-to-the-space-infidels/) [debate](http://programmers.stackexchange.com/q/57).

However, I think [this post](http://jarrodoverson.com/blog/spaces-vs-tabs/) explains the issue I see with tabs quite well. If tabs *would* have a globally defined width, there would be no issue, but they don't.
Tabs do work for indentation, but they don't work for alignment:
```js
// beginning of line
var one = 1,
    two = 2;

// end of line
var foo = {
  one: 1,     // one comment
  two: 2,     // another comment
  three: 2    // a third comment
}
```
This issue can only be solved by mixing tabs and spaces, which is completely confusing and complicated.
Thus I think spaces actually win over tabs in regards of readability, because it always looks exactly how it was written, anywhere.

----

> now we can move on and have some fun with the code ! 

Well, currently it's all set to 2 spaces, I guess we should decide on something first :)

If you'd prefer to use 4 (or any number of) spaces instead, I'd be more happy to use that instead of tabs.

PR #13, opened 2015-05-26
Music Library Updates

ning-yu:
Changes to these files were in progress when the eslint change took effect. Will run eslint on these in a separate commit.


volumio:
We should think on how\where to export the mpd and spop controllers...

ning-yu:
Will do.

PR #15, opened 2015-06-01
Add MPD Library Scan

ning-yu:


controllers/controller-mpd.js
```diff
@@ -65,254 +72,303 @@ function ControllerMpd(nHost, nPort, commandRouter) {
...
+			var objNewMetadata = {'title': '', 'artists': [], 'album': '', 'genres': []};
+			curEntry = {'service': 'mpd', 'uri': arrayLineParts[1], 'metadata': objNewMetadata};
+			objReturn.playlists.push(curEntry);
+		} else if (arrayLineParts[0] === 'Time') {
+			curEntry.metadata.duration = arrayLineParts[1];
+		} else if (arrayLineParts[0] === 'Title') {
+			curEntry.metadata.title = arrayLineParts[1];
+		} else if (arrayLineParts[0] === 'Artist') {
+			curEntry.metadata.artists = libFast.map(arrayLineParts[1].split(','), function (sArtist) {
+				// TODO - parse other options in artist string, such as "feat."
```
jotak: I think the "AlbumArtist" tag should also be read to be added to the artist list

volumio: Yes I think as well!

ning-yu: Good idea! What do you think about storing the MPD "AlbumArtist" as "performer" on Volumio? That would conform to the Ogg Vorbis standard that I've been using so far.

volumio: Agree, also flac lists:
8 - Artist/performer
https://xiph.org/flac/format.html#metadata_block_data
Date: Mon, 1 Jun 2015 11:02:29 -0700

PR #45, opened 2015-08-26
Added toggle command

hokilpet:
Asked for help here, but suggested to fix by myself
https://volumio.org/forum/toggle-between-play-and-pause-from-http-request-t3385.html

Is there any sense?

### DIFF
--- a/app/statemachine.js
+++ b/app/statemachine.js
@@ -83,6 +83,32 @@ CoreStateMachine.prototype.play = function(promisedResponse) {
 	}
 };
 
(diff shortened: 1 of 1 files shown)


ning-yu:
Hey Hokilpet,
It looks like you would like to add a toggle feature to Volumio 1.5? If so, you're probably looking for [this repo](https://github.com/volumio/Volumio-WebUI)!

hokilpet:
Thanks a lot!!

It is definitely what i searched for.

PR #67, opened 2015-10-07
Added GNU3 license

nickcannariato:
Noticed that the website had the license posted, but the repo didn't. Added the matching license to the repo.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
+  The GNU General Public License is a free, copyleft license for
+software and other kinds of works.
+
+  The licenses for most software and other practical works are designed
+to take away your freedom to share and change the works.  By contrast,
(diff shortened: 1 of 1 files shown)


volumio:
I'm delaying the decision on the Licence, since I'm not sure to stick with it or change to MIT or others. Open for suggestions

nickcannariato:
I'm personally a big fan of MIT because it's the simplest and most permissive (outside of The Unlicense, which carries no requirements at all). GNU is fairly stringent by comparison and is meant to keep the source code open. 

I guess ultimately it depends on whether/how you would like independent or commercial developers to use the source code in their own projects, should desire to do that at all. Thoughts?

jomo:
IANAL, but I think you need every [contributor](https://github.com/volumio/Volumio2/graphs/contributors) to agree on license changes. I'm not sure how the legal situation is when no License was present and a new one is added, but I guess the same would apply.

> I'm not sure to stick with it or change to MIT or others

I prefer MIT for most of my own projects but I'd be happy with any [Open Source License](https://opensource.org/licenses) as well.

volumio:
An open discussion with volumio's community  is here
https://volumio.org/forum/volumio2-licence-change-t3522.html

volumio:
I think we can go with GPLv3 for the system. Any contributor which doesn't agree please let us know, not telling anything before the licence change will imply agreeing on this

philkunz:
any updates here?

apinner:
@volumio are we any closer to choosing a license?

PR #95, opened 2015-11-19
restore UIConfig.json

gkkpch:
The UIConfig.json for plugin networkfs seems to have gone lost during merging


volumio:
Is not an error, there's a new Ui to support it

PR #122, opened 2015-12-13
Migrating switch to map

piercer:
Hi Volumio 2 team. I am an experienced developer who is also an audiophile. 

I use Volumio 1 on a raspberry Pi with my Almarro A340 valve monoblocks and a Chord 2Qute DAC. It works very well in a high-end system :-)

I am on the volumjo forums as piercer and am having difficulty getting set up for development. I do not have a spare Pi and so am trying to get development working on a VM following the instructions on GitHub without any success :-(

Anyway, I really want to help on this project - it combines 2 of my great passions.

Sp by way of introduction here is a small change to the InterfaceMPD. Large switch statements are inefficient and can be difficult to maintain. Therefore in this pull request I suggest using a command->handler map instead.

It also fixes one possible bug 

I think

case command.URLHANDLERS :
		self.handleUrlself.handlers(sCommand, sParam, socket);

Should be

case command.URLHANDLERS :
		self.handleUrlhandlers(sCommand, sParam, socket);

and this is reflected in my map replacement

Also the results variable in the while loop of the protocol server is actually global. I have put a var results in to fix this.

Please let me know what you think.

### DIFF
--- a/app/plugins/user_interfaces/mpdemulation/index.js
+++ b/app/plugins/user_interfaces/mpdemulation/index.js
@@ -140,6 +140,7 @@ function InterfaceMPD( context) {
 				return; // our buffer has received no full line yet
 			}
 
+			var results;
 			// while we still have a complete line in our buffer (os.EOL == end of line (\r\n))
 			while (results = buffer.split(/\r?\n/)) {
 				// get 1 line from our buffer to process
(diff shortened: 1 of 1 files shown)


piercer:
OOPS, forgot to mention - I have not tested this as I do not have a working dev environment. It was quite easy to see it is functionally equivalent though (I hope)

volumio:
Hi Piercer, it seems that you can be a great addition to the project! I'll help you setting up the dev environment (and we're rebuilding it as of now...).
I can't test your addition as of now (also because the mpdinterface is temporarily disabled...), but I'll do soon.
Appreciate this commit.
Get in touch with me with any means you prefer and I'll help with set up the dev env

Issue #136, opened 2016-01-06
webradio


balbuze:
There is something wrong when sorting webradio by countries : for instance if you choose "France" it return nothing. Uk and Italy are ok. Didn't test them all

gkkpch:
Yes, this is strange. When I do an manuel 
"http://api.dirble.com/v2/countries/FR/stations?token=51cc13037f15d06e727b34fe02351caea896f357&page=1&per_page=30" in the browser, 
it returns me a list of french stations (the token is in the dirble config.json)
Have not checked it in detail, could it be a codepage thingy?

balbuze:
yes <i was thinking about a codepage issue. I'm trying to test it.

volumio:
I confirm I get the problem as well, thanks for spotting this... If you guys also could debug it, it would be fantastic...
I'm preparing the new build in the meanwhile... Optimizing a lot for old raspis...

balbuze:
With GB codepage, same problem..Don't know if it help, but if you select France, nothing happens, but if you wait you can hear the startup sound trying to play 1 sec and on and on

balbuze:
The only way to stop it is to killall node

gkkpch:
Ok, probably I can help from tomorrow night. Played around with dirble last weekend, the gathered debugging knowhow might help. 
There is another issue with lwebradio, like paging while listing statons. Currently you only get one page (the first) with a max. of 30 stations. This is not a bug, paging does not seem to be implemented yet. Though for a beta it might not be top priority, it does show the principle.

mpasteur:
datas returned by  dirble seems fine for code FR (cf attachment) so it should be an problem during "decoding" these datas. I'll try to debug it soon...
[dirble_FR.txt](https://github.com/volumio/Volumio2/files/81745/dirble_FR.txt)
[dirble_IT.txt](https://github.com/volumio/Volumio2/files/81750/dirble_IT.txt)

gkkpch:
Yes, already noted that (see 2nd post), any help in debugging appreciated.

Also noticed, that stations without streams (missing "streams" element) get listed, but obviously when you click them, they can't play. Suggestion: skip a station unless at least one stream url is specified.

mpasteur:
hmm maybe we don't work on the same version but on mine the bug is exactly here. Some stations don't have any stream so there is an exception [TypeError: Cannot read property 'stream' of undefined]. As there is no catch within this plugin the list doesn't show...

mpasteur:
[...]

PR #142, opened 2016-01-08
Don't add station without streams

mpasteur:
see issue #136 
we don't add stations without streams
Caution : this method need better exception processing !


volumio:
Think that should be also valid for genres listing...

Issue #104, opened 2015-11-26
NasMounts config.json missing 'Path' variable


ecto1a2003:
the 'Name' is used as a path leading to mounts that are multiple folders deep. for example 'e/music' is the path to the share but the mount folder is created with the same path

balbuze:
yep it's a known problem.  Name is used as path... you can edit config.json to correct it, waiting for a fix.

volumio:
Yep, I accept suggestion on how the input fields should be named, and how many of them ...

balbuze:
1 - There is still a problem mounting nfs share : name is used as path. 
2 -And it is not possible to mount the same folder once with cifs, once with nfs (to compare cpu use because cifs seems use up to 3% with a rpi2) because the mount path cannot be define properly ( trough webUI)

mpasteur:
For me we should have : 
 * an host (IP adress or hostname) eg : 192.168.0.10 or mynas
 * a path which is relative to the host : mp3/classical
 * a name - fixed by user to easily find their shares : NAS1, NAS2 ...
 * maybe a login /password 

did i miss something ? i can work on this this wek-end i think.. (mostly with NFS i don't know if i can mount  a cifs share to test)

balbuze:
With the last built (from today 12/01/16) I think that network drive configuration is not clear. With cifs files, it's rather easy even if the name of the share and the path should be separate. With nfs same thing but path is much more complicated.

balbuze:
Mpasteur I read your comment after I post...

mpasteur:
no problem. If you have some time to test my PR i would be glad to have some feedbacks...

balbuze:
I'll try tonight. But I see a lot of activities !

balbuze:
@mpasteur 
There is something I don't understand, with your last PR I see no difference. Maybe I miss something but I thought you add field to configure as you wrote above

volumio:
You need to update both the UI and the Backend, plus you need to delete /data/configuration and reboot

balbuze:
Hello,
I have tested the PR from mpasteur. For me, it's good, understandable settings etc. Some minor improvements : window is not refreshed when adding or removing share. We should disallow the use of space in name as it use to create the mount point and does not work.

ecto1a2003:
i think im running the latest build and i dont see an option in the ui for path, how do i update both the backend and the ui.

ecto1a2003:
... and it just magically fixed itself. weird?

[...]

PR #144, opened 2016-01-09
Add "path" variable for mounts and other fixes for cifs and nfs mounts

mpasteur:
many changes to add a name to mounted path. Fixes for nfs and cifs with authentication
see #104 
Some changes in another commit on Volumio UI to add additional fields

Caution : nfs with authentication is not allowed at the moment....

### DIFF
--- a/app/plugins/system_controllers/networkfs/index.js
+++ b/app/plugins/system_controllers/networkfs/index.js
@@ -162,43 +162,39 @@ ControllerNetworkfs.prototype.mountShare = function (shareid) {
 	var defer=libQ.defer();
 
 	var sharename= config.get('NasMounts.'+shareid+'.name');
-	var pointer= '//' + config.get('NasMounts.'+shareid+'.ip') + '/' + config.get('NasMounts.'+shareid+'.name');
+	var fstype = config.get('NasMounts.'+shareid+'.fstype');
+	var pointer;
+	var fsopts;
+	var credentials;
+	
+	if( fstype=="cifs" ){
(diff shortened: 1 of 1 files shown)


volumio:
Thanks!!! I'll test and merge! Agree with you that this add way more clarity..

volumio:
Unfortunately on my setup that doesn't work...

My music is @ 192.168.1.142/Flac

inputting:
name: Flac
Path: Flac
IP: 192.168.1.142

And I don't get any alert..

mpasteur:
cifs or nfs ?

is your config correctly saved in config.json ?

volumio:
cifs... I'm diggin into that..
The conf gets saved, and the system reports as not mounted (the folder is created with the right name but its empty)

mpasteur:
ok don't know if it's the same bug as yours but it seems there is a problem
with "add a new drive"
If i add i get one message
info: [1452542003814] Error Mounting Share synomms: exec error Error:
Command failed: /bin/sh -c /usr/bin/sudo /bin/mount -t cifs -o
username=,password=,dir_mode=0777,file_mode=0666,iocharset=utf8
//synomms/music/mp3 /mnt/NAS/synomms
username specified with no parameter

username and password are empty but should be "guest" ( mount -t cifs -o
guest,dir_mode=0777,file_mode=0666,iocharset=utf8 //192.168.0.10/music/mp3
/mnt/NAS/synomms)

if i return to home , then come back to "My Music" and make an edit it's ok.

mpasteur:
i've made another PR see #146

PR #146, opened 2016-01-11
Nfs mount (v2)

mpasteur:
can you try this one ? there was a bug when adding a new drive with an empty username


volumio:
See my comments here: https://github.com/volumio/Volumio2-UI/pull/11

Issue #162, opened 2016-01-17
Radio does not sort stations alphabetically


atcapollo:
Radio stations are not sorted alphabetically. Under the country listing the UK is in the 'G' section.

The dirble api used to get radio stations doesn't sort. It has to be done manually.

volumio:
We have new API key with extended API requests.. please update the system to get it...

PR #167, opened 2016-01-18
issue #162 : sort radio stations and countries alphabeticaly

mpasteur:
list of countries and radio stations by country are now sorted alphabetically
(Pagination doesn't work atm...)


volumio:
Can't check it, since it seems that our API key is blocked...

volumio:
Ok, tested it. What do you think then would be the next step for showin all stations? 
(and sort also by genres type of radios?)

PR #176, opened 2016-01-22
Alsa Output Control Pt 1

piercer:
When searching through the sound card directory (/proc/asound/) the sound card numbers no longer need to be sequential. All cards with directories of the form card(\d+) and pcm0p/info files inside will be made available as Alsa cards.

This is the first stage of getting the selection of the output device working.

The next stage is to make sure that the selection is stored in the plugin configuration and that the MPD config file is updated accordingly - at the moment it is hardwired to hw: 0,0.


piercer:
Going to close this one and send a complete one - Still getting used to pull requests.

PR #181, opened 2016-01-23
Alsa Output Control From the UI

piercer:
**Functionality**
When searching through the sound card directory (/proc/asound/) the sound card numbers no longer need to be sequential. All cards with directories of the form card(\d+) and id files inside will be made available as Alsa cards.

The Audio device names are shown in a human readable form - coming from the card<>/id file instead of the card<>/pcm0p/infor file. this means instead of multiple 'USB Audio' entries in the UI you can actually see what USB audio device is being referred to.

After saving the new output device in the UI the mpd.conf file is updated.

After saving the new mpd.conf file an attempt is made to restart mpd and use the new config.

**File Permissions**
In the current distribution of volumio the mpd.conf file is read-write for user:mpd and read-only for group:audio. This means that user:volumio cannot update the mpd.conf file unless you change the permissions for the mpd.conf file by adding group write permissions

sudo chmod g+w /etc/mpd.conf

I do not know if thias is what is wanted, nor how to update it in the distribution so that it works out of the box.

Additionally the /etc directory does not allow anybody to write new things to it and so the /tmp directory is now used for the mpd.conf.old backup - which is actually the correct directory for such things anyway.

**Access Permissions**
The attempt to restart mpd does not work for the volumio user as they do not seem to have permission to do it. Again this is not a source code issue and so I am sending you this pull request. I am going to look into allowing user:volumio to restart mpd, and let you know if I find  way.

**this and self**
There seems to be some confusion in the volumio source code about the usage of self (apologies if there are not).

A 'self' reference to 'this' only needs to be created in an outer scope when it is to be passed into a callback function with an inner scope that will be resolved later. 

IMHO It is actually quite confusing to read all the 'self' references when 'this' would be correct.
IMHO Only using 'self' when it is needed makes the code easier to understand.

In the spirit of this I have change the alsa contrroller index.js to only use 'self' where it is actually needed. I think the code is cleaner, simpler and easier to understand (with less lines :-) )

What are your thoughts?

### DIFF
--- a/app/plugins/audio_interfaces/alsa_controller/index.js
+++ b/app/plugins/audio_interfaces/alsa_controller/index.js
@@ -1,30 +1,22 @@
 var io = require('socket.io-client');
-var libQ = require('kew');
-var exec = require('child_process').exec;
 var fs = require('fs-extra');
-var S = require('string');
 
 // Define the ControllerMpd class
 module.exports = ControllerAlsa;
 function ControllerAlsa(context) {
-	// This fixed variable will let us refer to 'this' object at deeper scopes
-	var self = this;
-	self.context = context;
-	self.commandRouter = self.context.coreCommand;
-	self.logger = self.context.logger;
-
+	this.context = context;
+	this.commandRouter = this.context.coreCommand;
+	this.logger = this.context.logger;
 }
 
 ControllerAlsa.prototype.onVolumioStart = function () {
-	var self = this;
 
-	var configFile = self.commandRouter.pluginManager.getConfigurationFile(self.context, 'config.json');
+	var configFile = this.commandRouter.pluginManager.getConfigurationFile(this.context, 'config.json');
 
-	self.config = new (require('v-conf'))();
-	self.config.loadFile(configFile);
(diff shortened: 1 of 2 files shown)


volumio:
Well done! 
I fixed the two issues you reported, with
https://github.com/volumio/Build/commit/22bd8aa23523034299c89016d082660233dc6d85
And by applying sudo to systemctl (we have permissions on this one)...

We're definetely going into the right direction! I'll detail what my ideas are on the next steps...

jomo:
> The attempt to restart mpd does not work for the volumio user as they do not seem to have permission to do it. Again this is not a source code issue and so I am sending you this pull request. I am going to look into allowing user:volumio to restart mpd, and let you know if I find way.

You can add this to `/etc/sudoers` before the group defintions:
```
root    ALL=(ALL:ALL) ALL
volumio ALL=(root)    NOPASSWD: /sbin/start mpd, /sbin/stop mpd, /sbin/restart mpd, /sbin/reload mpd
```
Now you can use `sudo restart mpd` as user volumio without a password.

volumio:
I already fixed it in the build script...

PR #187, opened 2016-01-25
MPD service optimisation

piercer:
Hi,

here are the first set of optimisation I have done to the MPD service. I have got it currently as fast as it can be with the current way it works. I think it is a bit faster.

I'm a bit tired now, so I'll leave it to you to try out.

Let me know how you get on.


volumio:
Well done!

PR #195, opened 2016-01-30
Using native bind method

piercer:
A few things - even though I seem to be up to date with the master there are changes that have appeared in this pull request that I did not do - in airplay_emulation, the www link and package.json.

The imprtant change in this pull request are to the websocket module.

1) I have made sure that the selfConnectWebSocket is not global as this seems like a very bad idea :-)
2) Under the hood the libFast library checks to see if the native bind function exists and then uses it if it does. We know it does already (Node 5.5.0+V8) and so we can use it without going through the libFast library first. If you like this I will do this wherever libFast is used.

Beyond this I am looking at moving to the native promise library that is in Node 5.5.0 from the V8 engine - that way we start getting Volumio towards as few extra libraries as possible. We will beneift directly from any work done on the V8 engine then :-)

Let me know what you  think.

I am running this code at home on the HiFi and it seems to work, but not fully tested.

### DIFF
--- a/app/.gitignore
+++ b/app/.gitignore
@@ -1,1 +1,2 @@
 testMusicLibrary.js
+Volumio2.iml
--- a/app/index.js
+++ b/app/index.js
@@ -467,3 +467,12 @@ CoreCommandRouter.prototype.reboot = function() {
 		} else self.pushConsoleMessage('Rebooting');
 	});
 }
+
+CoreCommandRouter.prototype.emitFavourites = function(msg) {
+    var self = this;
+
+    var plugin=self.pluginManager.getPlugin('user_interface','websocket');
+    plugin.emitFavourites(msg);
+}
+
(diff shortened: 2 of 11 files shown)


volumio:
Clever points... I'm testing this..
One thing: libfast was used to allow faster execution when compared to native bindings... Do you know if V8 native methods are faster than libfast?

volumio:
Nevermind doing a new PR, I merged and fixed the regression... Really good work, thanks!

piercer:
I tried a fetch upstream/master before making the pull request and everything was apparently up to date - which is why I was confused. Sorry.

piercer:
WRT performance - libfast uses the native binding method if its there - so it couldn't be any faster :-)

volumio:
Yep, at first I didn't get it...

PR #198, opened 2016-02-01
Play Queue Simplification?

piercer:
I noticed that in CorePlayQueue constructor there was a promise created for the play queue which is immediately resolved. 

```
       var self = this;

	self.commandRouter = commandRouter;
	self.stateMachine = stateMachine;

	self.queueReadyDeferred = libQ.defer();
	self.queueReady = self.queueReadyDeferred.promise;

	self.arrayQueue = [];

	self.queueReadyDeferred.resolve();

```
This promise was then returned every time that the play queue was accessed (adding, removing items and retrieving).

The add and remove items in particular wrapped another promise in this pre-resolved promise

```
return self.queueReady
		.then(function() {
			self.arrayQueue.splice(nIndex, 1);
			return self.commandRouter.volumioPushQueue(self.arrayQueue);
		});
```

This seems entirely unnecessary.

I have removed this pre-resolved promise and made the getQueue method syncronous. I also return the volumioPushQueue promises directly.

This all works and has had the side effect of fixing an issue where albums being added to the queue would show up multiple time (but I don't know how).

Now, it is possible that the pre-resolved promise is there for some future reason that I am unaware of, and so you may not want to do this. I think it is better and simpler now, and it has a side effect of fixing the issue mentioned.

Let me know

### DIFF
--- a/app/playqueue.js
+++ b/app/playqueue.js
@@ -1,56 +1,41 @@
+'use strict';
+
 var libQ = require('kew');
 var libFast = require('fast.js');
 
 // Define the CorePlayQueue class
 module.exports = CorePlayQueue;
 function CorePlayQueue(commandRouter, stateMachine) {
-	var self = this;
-
-	self.commandRouter = commandRouter;
-	self.stateMachine = stateMachine;
-
-	self.queueReadyDeferred = libQ.defer();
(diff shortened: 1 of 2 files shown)


volumio:
Testing now! Good point however...

volumio:
Seems to work great! Excellent!

PR #211, opened 2016-02-06
Tidying up index.js

piercer:
Hi,

I am starting to commit my code experiments :-)

In this one I simply tidy up index.js

it now uses strict mode, which can help V8 with optimisation.

I have removed a load of unnecessary calls to '.cal(...)' as the scope is not in question in these cases - when you call a method from an object that was created with new() then its scope is a bit more solid.

I have made some inroads with removing the internal unnecessary promises and pretend asynchronous code. these will start to come next.

I hope this is OK

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,4 +31,4 @@ node_modules
 *.swp
 spotify-login.txt
 .idea
-.iml
+*.iml
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,4 @@ node_modules
 *.swp
 spotify-login.txt
 .idea
+.iml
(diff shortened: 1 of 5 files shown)


piercer:
I have added some extra work to this pull request.

The statemachine has much simplified promise handling

For example, Instead of 

```
return self.updateTrackBlock()
				.then(libFast.bind(self.getState, self))		
				.then(libFast.bind(self.pushState, self));
```

there is

```
self.updateTrackBlock();
return self.pushState()
```

This has been done by allowing some of the inner workings of the state machine and playqueue to become synchronous.

Other code has been tidied up too.

This should probably all be merged before other committers start deviating too much

piercer:
Too much has gone on since I submitted this - merging again

PR #220, opened 2016-02-09
Strict!

piercer:
I added 'use strict' at the top of every js file. This helps with optimisation and actually revealed about 6 places where global variables were being created, some even with the same name - I fixed them.

Added semi-colons to the end of prototype function definitions.

Made a few array loops proper indexed loops - Interestingly, if you ever treat an indexed array as an associative array in V8 you lose all opportunity for it to be optimised, so this is quite important. 

This should all be functionally equivalent, just a bit tidied up.

### DIFF
--- a/app/metadatacache.js
+++ b/app/metadatacache.js
@@ -1,3 +1,5 @@
+'use strict';
+
 var libQ = require('kew');
 var libFast = require('fast.js');
 var libLevel = require('level');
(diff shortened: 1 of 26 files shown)


volumio:
Wow that's a great rework piercer! Anyway, from my tests, it breaks quite a lot of the system (since it reveals all "small errors" done), mainly we have lots of undeclared variables. My thought is that we should review (and fix) all those cases via intensive testing and then merge. 
Would you be available to do some testing\fixing and then edit the PR accordingly?

volumio:
Let me tell you that: it was quite useful to fix those strict mode errors, I feel like it will teach me a lot. And in the long term it will make the project better... thanks
Still, we need to test the system extensively to find other new errors, please help me with that...

piercer:
Really sorry about the errors you had to fix. I will not do it again without such rigorous testing. I am glad you found it useful though.

piercer:
I will do some tests tonight - am at work at the moment

PR #221, opened 2016-02-10
Adding directories to playlist

piercer:
I have made it possible to add a directory of songs to a playlist. In the current release it actually causes Volumio to reboot when you look at some playlists created this way - I do not understand fully, but this new version is better.

It does not yet work for directories inside directories yet, but I will work on that next.

In order to make this work in any sane manner I had to make the fake asynchronous method

`ControllerMpd.prototype.getAlbumArt`

synchronous - I do not know why it was asynchronous, it just returns a string that is calculated internally - was there a reason?

The whole album art setup is very confusing to me, and sometime it doesn't not seem to work, but I think this addition is a good place to start.

BTW we seem to be getting code in these things that only works for certain plugins (like if (service=='dirble') in mp/index.js which I had to merge with. I think we need a better way to deal with this, but it may be difficult.

Anyway I hope this is useful.

### DIFF
--- a/app/playlistManager.js
+++ b/app/playlistManager.js
@@ -73,21 +73,14 @@ PlaylistManager.prototype.deletePlaylist = function (name) {
 };
 
 PlaylistManager.prototype.listPlaylist = function () {
-	var self = this;
 
 	var defer = libQ.defer();
 
-	self.commandRouter.pushConsoleMessage('[' + Date.now() + '] ' + 'Listing playlists');
+	this.commandRouter.pushConsoleMessage('[' + Date.now() + '] ' + 'Listing playlists');
 
-	var playlists = [];
-
-	var folderContents = fs.readdirSync(self.playlistFolder);
-	for (var j in folderContents) {
-		var fileName = folderContents[j];
-		playlists.push(fileName);
-	}
-
-	defer.resolve(playlists);
+	fs.readdir(this.playlistFolder, function(err,folderContents) {
+		defer.resolve(folderContents);
+	});
 
 	return defer.promise;
 };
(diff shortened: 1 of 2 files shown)


volumio:
I think adding a whole folder could be useful, however we need to patch also the UI to do this...
Regarding the albumart, we discovered it adds a lot of overhead to the system (particularly after https://github.com/volumio/Volumio2/commit/0b8894a66eb65afc7abc20ced58148794625252b, since we look for id3 tag images, and therefore we need to retrieve every single file and scan it...).
So the plan is to make it more efficient, and add the option (which we have only for albumarts retrieved from web) to serve the image in different sizes (small, medium, large) by putting this as a parameter in the query (basically moving /extralarge at the end of the request URL).
Do you have an idea on how to make this efficient?

piercer:
I'm not sure why you say you have to patch the UI to do this, from the head of the maser UI branch of Volumio2-UI there is always an 'Add to playlist' option for directories - I thought it was intended. Should I not be using that version of the UI?

I think the change to make 'getAlbumArt' not return a promise is a good one - I now also have ControllerMpd.prototype.parseState and ControllerMpd.prototype.parseTrackInfo not returning promises - this makes the code that uses them simpler and we only have promises for things that are properly asynchronous, like sendMpdCommand etc...

How should I proceed?

piercer:
P.S. I'll think about albumart

piercer:
Also the new code works when adding single songs too - so IMHO there's no harm in merging it for now.

piercer:
I have merged in your latest work, and more more of the internals of mpd/index.js use synchronous code - the execution of the code was always synchronous, but it was hidden behind a promise :-)

piercer:
I have added an update to 'listPlaylist' in playlist manager properly asynchronous, which surprisingly it wasn't :-)

volumio:
Ouch I was in the process of merging...

volumio:
Coud you just send a PR for the last thing you changed? I merged the rest...

piercer:
Done - need to close this one first

piercer:
Nope - can't seem to get out of this one?

piercer:
OK, I think this pull request now has my latest stuff in it.

I fetched upstream, merged it in and repushed.

Hope this is OK

Everything seems to work when I test it.

Issue #32, opened 2015-08-14
Output Stack


jomo:
The basic idea is to support multiple/different output devices, e.g. send to AirPlay speakers.
See [Output Stack](https://github.com/volumio/Volumio2/wiki#output-stack) on the wiki.
I actually planned to figure out and contribute this myself but then was too busy for it, so I'll just share my thoughts with you.

The idea is that each audio output implements a common interface – let's call it `AudioOutput`.
Every `AudioOutput` is initialized with an audio source and has a few very basic functions to `enable`, `disable`, `setVolume`, …

Volumio would output sound to a raw PCM stream, which is then passed as the audio source to all the outputs.
An `OutputController` enables or disables outputs and also calls `setVolume` for each of them.

Some output examples:
- take the PCM stream and send it to local speakers via DAC
- use [node_airtunes](https://github.com/lperrin/node_airtunes#playback) to send the audio to AirPlay speakers.
- convert to mp3 stream and make it accessible via HTTP

AirPlay is a great example because it shows two issues that I don't yet know how to solve:

**Synchronizing**
Especially remote speakers will have some delay before actually playing the audio, and you don't want your AirPlay speakers 5 meters away to hang 3 seconds behind the directly connected speakers.

*AFAIK AirPlay buffers audio and plays it back delayed, while pause/resume is carried out as soon as it arrives at the AirPlay server. However, that issue could be solved by not using AirPlay's pause/resume feature and just silence the PCM stream when Volumio is paused. This will have the effect that music still plays for a few seconds after you hit pause, though.*

Maybe `enable` could have a callback that includes the delay and `OutputController` can delay the source individually for each output to counter the latency?

**Custom Options**
Since you can select multiple AirPlay speakers, we need a way to pass this option to the end user.
Same goes for e.g. port options for the http server mentioned above.

Perhaps `AudioOutput` could have `settings` and a function that's called when they change?

----

I hope this made some sense :cake:

volumio:
It does, we just prepared a plugin system with context. Want to jump in on this and at least validate the system and give feedbacks?

PR #226, opened 2016-02-13
Output Device Change Listening

piercer:
OK, I want to get airplay working - my wife complains that it is currently broken in version 2 - which it is.

Problem 1) the airplay emulation (why is it emulation) plugin does not respond to changes in out put device.

Solution: Add the concept of output device change listeners to CoreCommandRouter. Make the airplay emulation plugin register as a listener and respond to changes when they occur - done!

Query: I have always thought it odd that the mpd plugin is the main controller for the playback setting UI. Maybe this could be changed to a more centralised playback settings system now that we have this system and simply broadcast changes to the listeners - that way other plugins can easily pick up on these changes. Anyway I have also made ControllerMpd add itself as a listener for output device changes and respond to them - even though it also generates the change events for now.

Result) the airplay emulator now outputs sound through the selected output device.

Problem) The bitrate of the sound played is completely wrong and it sounds all electronically garbled - PLEASE can you point me in the right direction to fix this (FYI Rune ausio still has this issue and it was my main reason for switching to Volumio 1.55 which did not).

Anyway - hope this is helpful.


piercer:
Merged in the latest updates - no conflicts

volumio:
Hi Piercer, I'm taking my time to see the PR... Since we already have a callback method for when the name changes, it would be reasonable to use it also for the playback device... 
I'll write more about it today and let you know...

piercer:
Hi Michelangelo, I have had a look and I think you are right, the callback system is more versatile than my one function handlers and would work for the output device changes.

I have two comments though.

1) The callbacks are really callbacks, they system you seem to have made is an even handling system: the difference  is like this

callback: Do something and tell me when you have finished (call me back)
event handling: Tell me whenever the thing I am interested in happens

So I think renaming callbacks to eventHandlers would be clearer.

2) In strongly typed languages (like java for example), events are normally actually objects, with a type defined. In Javascript we do not have that luxury and so the type (you called it 'name') of the event is a string. With the name change I have suggested above the way to register for an event would be something like this

`this.commandRouter.addEventListener('outputDeviceChanged', this.onOutputDeviceChanged.bind(this));
`

And when the output device is changed the commandRouter would call something like this

`this.fireEvent('outputDeviceChanged');
`

This is OK, but the code is 'stringly typed'

http://c2.com/cgi/wiki?StringlyTyped

It is hard to no be stringly typed in Javascript but it would be better if we could encapsulate the set of allowed events somehow so that spelling mistakes would not be the cause of bugs.

something like a global object like this (yes I know, a global :-))

```
const EVENT = {
  OUTPUT_DEVICE_CHANGED: {type: "outputDeviceChanged"},
  DEVICE_NAME_CHANGED: {"deviceNameChanged"}
};

this.commandRouter.addEventListener(EVENT.OUTPUT_DEVICE_CHANGED,this.onOutputDeviceChanged.bind(this));

this.fireEvent(EVENT.OUTPUT_DEVICE_CHANGED);
```

In code that used events the 'event.type' reference would actually give you the string name of the event.


What do you think?

piercer:
OK :-) Because we have a time delay between us, I have decided to take the bull by the horns and code up the proposed implementation above.

It makes the code look quite clean, forces you to at least have a 'type' for the event and I have provided a module of ready made events already in use.

If you don't like it, then never mind, but a system like this is great for cleanly decoupling the different parts of a system. In particular  the plugins no ;onger need to know about each other in order to communicate with each other. 

Let me know what you think.

volumio:
Dear Piercer, as for naming I think this is absolutely right... As calling them events would be more comprehensible...
I'm not really sure about the volumioEvents "dictionary" yet. Having the events stringled may remove a bit of complication from coding, at the price of possible mistakes...
Since I like your idea, I think we should take a little time and do a bit of a blueprint, in order to handle different scenarios with such method (for example a plugin registering\unregistering its playback capabilities).
I will create a wiki where we can outline this system, and once we cleared it we can then develop and merge. What do you think?

jomo:
> I have always thought it odd that the mpd plugin is the main controller for the playback setting UI. Maybe this could be changed to a more centralised playback settings system now that we have this system and simply broadcast changes to the listeners - that way other plugins can easily pick up on these changes

If I understand you correctly, you might want to take a look at #32 :)

piercer:
Hi,

its a bit late here and I can't comment too much, but I think you need to think about the problems of using strings - here is why :-)

You seem to have already defined three events

Push state.
Set volume.
Update volume.

The event string you defined for push state was 'volumioPushState'. How would anybody else know that string is meant to be, where would they look it up? 

Anyway forgetting that question for the moment, just say I want to listen for the set volume event, what string should I use? Based on the push sate event I would imagine 'volumioSetVolume' - seems logical doesn't it? Unfortunately I would be wrong, as it is defined as 'volumiosetvolume'. 

Update Volume? 'volumioUpdateVolume' or 'volumioupdatevolume' - hard to know seeing as the first two events had conflicting camel case standards. Where would I look to find the answer?

Do you see that even in defining just three event inconsistency and confusion has been created in their naming convention  - imagine if there are 10 or 20 events - very confusing.

Having a system that centralises event definition and removes any possibility of confusion over the string that is actually used gets around this cleanly.

piercer:
Any thoughts?

jomo:
I think the actual issue is that naming is inconsistent, switching from strings to objects doesn't really solve this problem, since you still need to get the correct variable name (where the event object is stored at). The 'Stringly Typed' issue is an interesting one, but isn't it [the standard](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) for JavaScript events?

I suggest definition a proper event naming convention and cleaning up those that don't match it.

fanciulli:
Hello @piercer ,

the callback thing has been put in volumio after a small hack i made in the volumiodiscovery plugin. That's why it has grown "wild and free".

I agree with you about the renaming but i'm not sure the stringly typed solution is the best for our needs. Here's why:

1) my idea was that plugins can register for a specific event through a clear method. This means for example that if i user the method registerDiscoveryUpdates i won't need any other parameter except the data on my callback. I exactly know what to expect.

2) Using the suggested implementation will consume cpu on checking if the callback is interested on the event before executing it. I know these are very few cycles but Volumio is currently in "Performance tuning" mode.

I'm anyway open to any comment or proposal. 
I would like to thank you all for the effort you put in Volumio. You guys are great! Even though you fix my own code... ;)   

Massimiliano

piercer:
My original implementation had a specific function to register for output device change events, but after talking to @volumio we went with the generic event listener system.

I have suggested making the events not be strings and have them listed in a central module - this prevents wrong event names and makes them easy to find.

WRT checking for an implemented callback - that doesn't really happen as the callback is passed through when registering - have a look at my pull request

Am at work, so this is all I can type for now. I am strongly against relying on people to correctly format strings.

@Volumio: How will future plugin developers know what events are available if there is no central place where they are listed? It helps the code document itself.

piercer:
Personally I think the following snippets

`this.commandRouter.addEventListener(events.OUTPUT_DEVICE_CHANGED, this.onOutputDeviceChanged.bind(this));
`

triggered by

`this.fireEvent(events.OUTPUT_DEVICE_CHANGED);`

and in the events module

`exports.OUTPUT_DEVICE_CHANGED = {type: "outputDeviceChanged"};`

is pretty clean code. Easy to read and understand.

fanciulli:
If you think about a future with a lot of plugins made by people i see this approach difficult to maintain, even though it is clean. Maybe we can use two distinct approaches for "internal" events and plugins events. What about this @volumio  ?

piercer:
@fanciulli - Which approach do you mean is difficult to maintain?

Normally in large systems the 'event bus' is a documented system. Plugin developers should be able to register for the advertised events that they are interested in and not register for the ones that they are not.

It is difficult to make javascript secure, but you can try to prevent plugin developers accessing the 'internals' by making this event system only accept the listed events, and not allowing them to use strings - not foolproof, but better.

The problem then becomes deciding which events to publicise :-)

fanciulli:
If you take this line:

this.commandRouter.addEventListener(events.OUTPUT_DEVICE_CHANGED, this.onOutputDeviceChanged.bind(this));

you see that you're associating the method onOutputDeviceChanged to the string defined by events.OUTPUT_DEVICE_CHANGED.

The more events are needed by third party plugins the more the events structure shall grow. If we want to give third party developer the freedom to create event emitting plugins we have to deal with a wide set of event strings. I think that this may lead to multiple plugins defining the same exact string. Imagine the mess.

I think the approach is good for "Volumio" deifned events. As you pointed out this approach is common in a lot of systems. I think we have to provide another mechanism (or leave developer freedom on this) for plugin defined events/callbacks

piercer:
Pligin developers could provide their own events module and the central system could be used for that too

```
var aPluginsEvents=require('APluginsEvents.js');
this.commandRouter.addEventListener(aPluginsEvents.SOMETHING_HAPPENED,this.handleSomething.bind(this));

```

Then 'APlugin' can fire the event 

`this.commandRouter.fireEvent(aPluginsEvents.SOMETHING_HAPPENED);
`

possibly...

piercer:
Actually - here might be the answer -

http://stackoverflow.com/questions/8595509/how-do-you-share-constants-in-nodejs-modules

 It'll give us the string constants that @Volumio likes, but encapsulated in a way that makes them constant and listed in one place.

I can code this up if you think it ticks all the boxes

fanciulli:
So you say that every plugin will export its events in a file. Let's say it is a sort of manifest for events....

THe only point we have to fix is how to handle two or more plugins using the same string for an event. Let's say PlaginA and PluginB declare the event "UPDATE".

I see the following options:

1) register will succeed and function have to be reliable when called. This anyway will break  one or more plugins at runtime.

2) Create a method for register the event so that the second plugin trying to register an event will fail.

3) Every event shall contain a plugin constant so that events are always different.

4) Use another mechanism.

Method 3 is the easiest and quicker to follow while the 4th is the safest.
Method 2 would help finding two plugins colliding but won't avoid having something not running on Volumio
Method 1 is the approach is won't follow at all. It is umpredictable..

frequencx:
Can we expect for this Idea to be merged in the long run? I have very much been looking forward to automatic Output Device listening!
What needs to happen?

volumio:
We've already implemented it... Look here 
https://github.com/volumio/Volumio2/blob/dev/app/plugins/music_service/mpd/index.js

line 758

PR #255, opened 2016-02-25
added icons

apinner:
Added icons to system plugins for use in UI headings.


apinner:
indentation was wrong, will resubmit.

PR #256, opened 2016-02-25
Plugin icons for settings headers

apinner:
Added icons for section headings for new UI


volumio:
That seems a very good idea! Give me some days (I'm abroad) and I'll merge!

volumio:
The default albumart is marvellous! It's what we were looking for! Testing now...

PR #262, opened 2016-02-29
Use standard date format

Alex131089:
https://en.wikipedia.org/wiki/Date_format_by_country, https://en.wikipedia.org/wiki/ISO_8601
I never heard of YYYY-dd-mm before :)


volumio:
That makes sense... I will also correct the image names

Alex131089:
Image names seems corerct from what I see :)

PR #277, opened 2016-03-13
Fixing Airplay distortion

piercer:
Hi Volumio

I have managed to get airplay working in Volumio2 without distortion through my DAC's.

All that is necessary is to change 'hw' to 'plughw' for the device definition.

My wife is now happy :-)

Conrad

### DIFF
--- a/app/plugins/music_services/airplay_emulation/index.js
+++ b/app/plugins/music_services/airplay_emulation/index.js
@@ -8,98 +8,68 @@ module.exports = AirPlayInterface;
 
 
 function AirPlayInterface(context) {
-	var self = this;
 	// Save a reference to the parent commandRouter
-	self.context = context;
-	self.commandRouter = self.context.coreCommand;
+	this.context = context;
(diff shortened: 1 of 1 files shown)


volumio:
Great find!

jomo:
[Here](https://www.mail-archive.com/alsa-user@lists.sourceforge.net/msg21124.html) is a great explanation of the difference between `hw` and `plughw`:

> plughw supports much more sample formats / channel configurations than underlying hardware supports natively, and performs conversion if needed. hw performs no conversion, but supports less configuration, sometimes only very obscure ones, but when used you can rest assured that no conversion takes place.
> 
> Mostly you'd want these conversions to take place, like mono->stereo conversion or S16LE->S32LE conversion etc (all depending on your hardware)
> 
> > If I need to drive the sound card directly should I use 'hw'?
> 
> You can use hw, but it isn't really recommended unless you REALLY know how to use it and have support of every weird format in all possible combinations. That's about underlying hardware details, and most applications don't want to deal with them.

PR #358, opened 2016-04-13
Fix submodule link

glucaci:


volumio:
I think we could also remove the submodule thing...

PR #378, opened 2016-04-22
Fix browseLibrary Uri, should use string lib.

glucaci:


volumio:
Why this commit? Was there any situation where uri was not a string?

glucaci:
If we don't use the string lib, will crash on musiclibrary line 652 at .startsWith().

glucaci:
In VM crashed, so I've tested on my RP and is no problem. Sorry

PR #466, opened 2016-06-17
Dev adding french language

balbuze:
I18N - French language added


volumio:
Wow!!! Thanks! That was fast!

volumio:
If you want to have even more fun... 

https://github.com/volumio/Volumio2-UI/tree/development/src/app/i18n

we got translations complete also in UI!!

PR #477, opened 2016-06-24
french language

balbuze:
New translations


volumio:
Please wait when I finish... there is lot to translate... you risk to do work and then redoit...

PR #546, opened 2016-07-12
Spanish Translation

mundodisco8:
Added the Spanish translation file and modified languajes.json

I wasn't able to create a build environment and test it thoroughly, but I think 99% of it it's OK. This would have been great in order to test it in mobile, to check for text overflows.

 Some of the constructed phrases (as "NAS was mounted successfully") might be in broken spanish and need a retouch. Also, there is one or two words that I couldn't find in the interface and I had to guess the context of their meaning, I hope I guessed right.


volumio:
Gracias!

mundodisco8:
Is the least that I can do after getting so much from the project. Thanks for your work, guys!

PR #549, opened 2016-07-14
Add Chinese translation

howardqiao:
Add Chinese translation


volumio:
Great!!! Many thanks!

PR #576, opened 2016-07-23
fixes #565 Wired connection shown even tough no cable is connected

### LINKED ISSUE #565: Wired connection shown even tough no cable is connected
wired cable
I only use WLAN still there alway is a wired connection shown with the same IP. No ethernet cable is connected.

Using RC2 hotfix.

overflowed:


volumio:
This is a consequence of a fallback IP we use when no cable is connected for diagnosis purposes... 
Let me check a bit what could be the best solution, yours is quite interesting

volumio:
Thank you for your PR, I solved it at its roots...

https://github.com/volumio/Build/commit/7a46234381e636dc500712d099e0756a0e94b7fd

Thanks!

Issue #599, opened 2016-07-29
After todays update of Volumio2: Can't reach neither web server nor ssh..


heis2201:
Installed update from 0.928 using the WebUI. Feedback was positive. After restart, can't reach web server or ssh server.
Tried to restart several times, to no avail. My hardware is Rpi B+ (headless, thus hard to diagnose if ssh doesn't work).

heliosphan79:
Same here... On my rpi3 everything worked out fine. my rpi2 isn't accessible anymore and kills the node process with an out of memory notice...

nosferatujr:
Could be the same happens to me, I thing that is caused for the language (see issue #598), really I reach the volumio trough SSH, and change the language, and after the restart the web server works again.  
But after back to English, the web server appears today off again ones, but after a new reboot all works again.
Something is happening, need investigation to found the real cause….

nosferatujr:
@heliosphan79 Can you ssh your rasp2?

skikirkwood:
In installed 0.976 on a Rpi3 with no problem, but can't connect to it in any way on a Rpi B+.  It boots up and my Rpi3 sees it with the multi-room interface, but I can't select it, and with the Rpi3 turned off I can't ssh into or get the Web interface for the B+.

nosferatujr:
but not sure, Really I reboot in english few times with success, is switch language to catalan, works until reboot. on the reboot ssh works and webserver not, changing /data/configuration/miscellanea/appearance/config-json to English the e initial sound is played and webserver works again...
Anyone of the guys that experiment the problem is using different language?

heliosphan79:
@nosferatujr, no ssh access... My rpi doesn't respond to anything... Hopefully you guys can find a quick fix... I'm missing the romantic music in my bedroom ;-)

Tihmann:
same problem here. Update and now volumio is not available anymore....

Tihmann:
I am using a Rpi 2

volumio:
So you did an update? From which version?  Is there any body  willing to reflash and see if that changes things?  No quick fix since I am in holydays...

balbuze:
I did a update from 0.975 and it results a kernel panic. A fresh install works on RPI2. Haven't test other devices.

JedS:
Same here.
Update from 0.928 and got a kernel panic too.
node process seems to take too much memory as I got an 
**Out of Memory Killing process 928 (node)
Killed process 928 (node) total-vm 336164Kb**
[...]

PR #617, opened 2016-08-01
Better Optimised Backgrounds

apinner:
Spent a bit of time trying to better optimise the background images to
help with https://github.com/volumio/Volumio2/issues/599

Reduced by 5MB in total, less than half of their original size.


heis2201:
How about having those background images as a totally optional thing. Users with plenty of RAM can use them, others on Rpi1 with only 512MB may just not use them. They are eyecandy, not an essential function.

apinner:
Do you really dislike our UI, you seem to want to remove all of it!? ;)

Nothing wrong with some eyecandy, we want volumio to look good! We will rework the backgrounds so they are optional.

@volumio as an idea, why don't we host the images and only include a thumbnail. Then when the user selects a background it downloads it behind the scenes?

heis2201:
@apinner No, don't get me wrong. I love the UI. I am just teasing you to first put effort on functionality and then on the eyecandy. 

I really appreciate all the effort that is put into the project by the team behind Volumio. A million thanks!

Does the standard volumio install come with a swapfile enabled? That could be an option for those devices with limited RAM, even though is slows down the OS, it wouldn't crash...

apinner:
Trouble using a swap on the SD card is you increase the read/writes and wear out the card. Volumio is designed to run entirely in RAM for that reason.

heis2201:
I can maybe manually create swap file on an (old) external 1gb usb thumb drive, where I don't care if it breaks... plug in the next one..

volumio:
Thanks alex!

PR #673, opened 2016-08-22
Add support for JustBoom boards

shawaj:
Add support for JustBoom boards


volumio:
I see a problem, you've got the same i2c address as Hifiberrry. So we won't have automatic recognition here... Do you see an alternate method we could use?

shawaj:
We could read the EEPROM data?

That has our product numbers in it.

I2C address can be the same with different boards so is not reliable way to
determine in my opinion.

Of course EEPROM way will only work with HAT versions with EEPROM for auto
detection, so with zero boards with no EEPROM it won't work.

Thanks

volumio:
Is there any doc to read the eeprom board?

shawaj:
When the Pi boots, in standard Raspbian, the EEPROM data is stored in
userland.

See this link here -
https://www.raspberrypi.org/forums/viewtopic.php?f=29&t=108134

To check if your HAT is recognized, just go to /proc/device-tree/. If you
see a hat directory, you are a winner:)

cd /proc/device-tree/hat/
more vendor
more product

shawaj:
Here is a typical output from that:

pi@raspberrypi:~$ ls -l /proc/device-tree/hat/
total 0
-r--r--r-- 1 root root  4 Mar 16 17:03 name
-r--r--r-- 1 root root 14 Mar 16 17:03 product
-r--r--r-- 1 root root  6 Mar 16 17:03 product_id
-r--r--r-- 1 root root  6 Mar 16 17:03 product_ver
-r--r--r-- 1 root root 36 Mar 16 17:03 uuid
-r--r--r-- 1 root root 20 Mar 16 17:03 vendor

pi@raspberrypi:/proc/device-tree/hat$ for f in *; do (cat "${f}"; echo); done
hat
RPi-Display B+
0x0001
0x0001
90a11443-7057-4587-8e56-fe7d5d1c7022
Watterott electronic

volumio:
Aaron, what's the /proc/device-tree/hat/product content that you have for any of your DACs?

shawaj:
/proc/device-tree/hat/product content for ours is:

JustBoom DAC HAT
JustBoom Digi HAT
JustBoom Amp HAT

It might be good to have it in Volumio so that you can declare multiple /proc/device-tree/hat/product names for the same driver/option in dacs.json because then you could handle situations like:

JustBoom DAC HAT v1.0
JustBoom DAC HAT v1.1 
etc

Whereby they use the same driver, but are different versions. What do you think?

Cheers

volumio:
Ok, that's a good one, I'll allow more product names..

shawaj:
Awesome :-)

volumio:
Added the eeprom with your suggestion here 
https://github.com/volumio/Volumio2/commit/29fb672a3fd52727cf52154bc632619a8b0d059d

Now building an image...

shawaj:
Looks great. So we no longer need to define i2c address... Just the EEPROM
dump name?

Thanks

shawaj:
Or both?

volumio:
Both, first we see if there's a proper EEPROM name, if we can't match it, then we use the i2c address as fallback...

volumio:
Here you have it:
http://updates.volumio.org/pi/volumio/0.981/volumio-0.981-2016-08-25-pi.img.zip

If our plans were correct, you should have your DACs  recognized automatically... 

Let me know how it goes!

shawaj:
Ok great, will try in the morning :-)

Do we need to declare an i2c address then in dacs.json? Is this i2c used
for anything else other than auto detection?

volumio:
I've updated the doc as well:
https://volumio.github.io/docs/I2S_DACs/Adding_Compatibility_to_your_DAC.html

So i2c is just a fallback, but the 4c address is already used so it cannot be entered. Let me know how it goes... Please let us know also for the other thing...

shawaj:
Ok great :-)

PR #692, opened 2016-09-08
Update of Swedish and English language files

TheRealMattiasJohansson:
Added Swedish translation for new items. Also did som tab fix (changed
from tab to blankspace).

There were som tabs instead of blank spaces in the English language file.


volumio:
Yes, it's meant on purpose

PR #729, opened 2016-09-28
Make the comments reflect the purpose of its functions

cgtobi:
The comments in the various functions were just plain copy and paste. This brings them in line with the actual purpose of its functions. More details might be needed.


volumio:
You're right... We will also change the example plugin to reflect the new used functions...

PR #752, opened 2016-10-10
Consistency in playlist API

gotling:
Change property to "name" to be consistent with rest of API and documentation. Updated in UI as well.


volumio:
Thank you!!! Coherence is always a good thing...

PR #765, opened 2016-10-13
Added missing Shebang line

SeproDE:
I added the missing header line in [playback.js](https://github.com/volumio/Volumio2/blob/master/app/plugins/system_controller/volumio_command_line_client/commands/playback.js), wich prevented commands like `volumio next` from working.


volumio:
Thanks!

PR #771, opened 2016-10-16
Fix undisplayed warning

xipmix:
Without the COMMON. prefix, the ILLEGAL_CHARACTER_/ warning is not displayed
by the warning popup.


volumio:
Thanks!

PR #772, opened 2016-10-16
Handle undefined or 'blank' share alias

xipmix:
If the user fails to type anything in the 'Alias' field of
network-drives-add-edit-drive.html then addShare fails with the following
somewhat obscure error in the logs:
    Missing error handler on `socket`.
    TypeError: Cannot read property 'indexOf' of undefined
and no indication to the user of what the problem is (particularly if this
is their first attempt to add a network drive).
In addition, the share is not mounted.

Fix by making sure 'name' is always defined and raising a warning in the webui
when 'name' is empty or all-spaces. To get a displayable warning I had to copy NETWORKFS.ALIAS_DOC from the webui into the string list.


volumio:
Very well done!!! thanks!

Issue #796, opened 2016-10-21
web radio lists not sorted alphabetically


jo-me:
the list of "My Web Radios" & "Radio Favourites" are currently sorted according to the time the items were added to the list.
It quickly becomes hard to find a specific entry even after adding only a handful of stations because you have to look through the whole list.

Also see https://volumio.org/forum/webradio-list-not-sorted-alphabetically-t4853.html

![image](https://cloud.githubusercontent.com/assets/10351272/19606168/d17e9674-97c0-11e6-82ee-3353759675a1.png)

PR #797, opened 2016-10-21
Sort webradio lists alphabetically

jo-me:
Fixes issue https://github.com/volumio/Volumio2/issues/796


volumio:
Very good thanks!

PR #869, opened 2016-11-22
Avoid NaN results when NAS path has spaces in it. Fixes #856.

### LINKED ISSUE #856: NAS size error: NaN when path has spaces
There seem to be a nasty bug in NAS shares size extraction [in here](https://github.com/volumio/Volumio2/blob/master/app/plugins/system_controller/networkfs/index.js#L523).
Indeed when share path has spaces in it, then `awk` gets mixed-up, and does not returns the proper column!...
Following size calculation is then altered and may end-up with NaN, which is what got displayed in UI.

xipmix:
Add quoting to the mountpoint path and avoid printing the NAS's
export path, so that we reliably get a raw number back from 'df'.
Add formatBytes() function, to give a size string suitable for
display in the UI.

### DIFF
--- a/app/plugins/system_controller/networkfs/index.js
+++ b/app/plugins/system_controller/networkfs/index.js
@@ -9,6 +9,27 @@ var mountutil = require('linux-mountutils');
 var libUUID = require('node-uuid');
 var S = require('string');
 
+// Library function based on
+// http://stackoverflow.com/questions/15900485/correct-way-to-convert-size-in-by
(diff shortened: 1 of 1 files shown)


xipmix:
I have no idea what I am doing in JS so treat with caution. In particular did I do the right thing with the stdout string? I did do some basic testing of formatBytes().

xipmix:
see also http://numeraljs.com/

xipmix:
Closing, this code did not even work. Sorry for the noise.

PR #872, opened 2016-11-23
Add Traditional Chinese translations for TW and HK based on zh

leafy7382:


volumio:
Thanks, is it different from the one we currently have? Also we need the locales for UI to be committed, see here: 
https://volumio.org/forum/contribute-translating-volumio-your-language-t4289.html

leafy7382:
Yes, the character set is different. Current zh only maps to Simplified Chinese used mainly in China and South-east Asia. The ones I added are for Traditional Chinese, mainly used in Taiwan and Hong Kong. Some terminologies have also been modified to local terms.

leafy7382:
Hi I have also submitted corresponding pull request to volumio2-UI for the language, could you please kindly incorporate both of them?

volumio:
Thanks!

PR #877, opened 2016-11-26
mount.cifs fails if given '//server//path'; so clean up the input path 

xipmix:
It seems best to just silently fix the mistake than bother the user with popups (that would have to be translated).


xipmix:
Hi, is there some reason not to merge this?

volumio:
Should we still merge this?

PR #883, opened 2016-11-27
Be consistent about checking the result of mountShare.

xipmix:
The 'success' case uses ==, the 'fail' case uses ===. Since the values are
coming from a function in the same module, there's little chance of them being
inconsistent types, so use === for success too.


xipmix:
When looking at this inconsistency I wondered about using == everywhere, as it would be sad to not respond to a 'fail' due to a type difference. 

Is it worth adding a final 'else' case that logs what the result was, if it wasn't 'success' or 'fail'?

PR #886, opened 2016-11-28
Don't allow volume to be stepped below zero. Fixes #857

### LINKED ISSUE #857: negative volume iin web UI
Volume control can be set to minus if you start at zero, then click plus (--> 12) then click minus 2 times (--> 12 --> 2 --> -8)
volume

xipmix:
I was able to reproduce this easily and the obvious fix seems to work. Not tested with a DAC, just onboard analog out.


volumio:
Well done!

PR #944, opened 2017-01-09
add driver=nl80211 in hostapd.conf

macmpi:
to be consistent with https://github.com/volumio/Build/pull/114
BTW it very much seems hostapd.tmpl in build  (https://github.com/macmpi/Build/blob/master/scripts/volumioconfig.sh#L419) does not serve much purpose.


volumio:
Thanks, that makes sense!

PR #949, opened 2017-01-11
Fix typo in commit #3a09b4b

macmpi:
Sorry, my bad (wrong cut&paste)


volumio:
Thanks for fixing&reporting!

PR #955, opened 2017-01-13
Adding pisound to dacs.json

gtrainavicius:
Adding pisound (http://blokas.io/pisound) to volumio.

Note that the latest pisound module code is available in 4.4.41 RPI kernel version.

Also I found that if switching from dac with 'needsreboot=yes' to 'needsreboot=no', reboot should still happen, without it, the output doesn't work.

### DIFF
--- a/app/plugins/system_controller/i2s_dacs/dacs.json
+++ b/app/plugins/system_controller/i2s_dacs/dacs.json
@@ -22,6 +22,7 @@
     {"id":"mamboberry-dac","name":"Mamboberry HiFi DAC+","overlay":"hifiberry-dac","alsanum":"1","mixer":"","modules":"","script":"","needsreboot":"yes"},
     {"id":"phat-dac","name":"pHAT DAC","overlay":"hifiberry-dac","alsanum":"1","mixer":"","modules":"","script":"","needsreboot":"yes"},
     {"id":"picade-hat","name":"Picade HAT","overlay":"hifiberry-dac","alsanum":"1","mixer":"","modules":"","script":"","needsreboot":"yes","eeprom_name":["Picade HAT"]},
+    {"id":"pisound","name":"pisound","overlay":"pisound","alsanum":"1","mixer":"","modules":"","script":"","needsreboot":"no"},
     {"id":"raspidacv3","name":"RaspiDACv3","overlay":"raspidac3","alsanum":"1","mixer":"Digital","modules":"","script":"","needsreboot":"yes"},
     {"id":"rpi-dac","name":"R-PI DAC","overlay":"rpi-dac","alsanum":"1","mixer":"Digital","modules":"","script":"","needsreboot":"yes"},
     {"id":"soekris-dac","name":"Soekris dam 1021","overlay":"hifiberry-dac","alsanum":"1","mixer":"","modules":"","script":"","needsreboot":"yes"},


xipmix:
Where you say:
> 'needsreboot=no', reboot should still happen

do you mean 'the system still needs to be rebooted when the module is loaded'? If so, why do you set 'needsreboot=no' ?

gtrainavicius:
If you have already selected a dac which had 'needsreboot=yes', switching to a DAC which wouldn't normally require reboot, still requires it, otherwise it won't work until rebooted. If switching from built-in audio to, everything works fine.

volumio:
@gtrainavicius unfortunately we need to wait for the new kernel before merging the PR...

gtrainavicius:
@volumio When is that planned to happen?

btw, about needsreboot, the clearest way to express what I meant is probably code :) I mean the condition whether to restart after switching the card should be:

`shouldReallyReboot = oldCard.needsreboot || newCard.needsreboot;`

macmpi:
> @volumio: we need to wait for the new kernel before merging the PR

Will new kernel still be on 4.4.x or bumb to [LTS 4.9.x that is getting ready](https://www.raspberrypi.org/forums/viewtopic.php?t=167934)?

gtrainavicius:
@volumio Any updates on when you plan to update your kernel?

volumio:
Hi, sorry we're still holding as the kernel updater is not ready yet... apologize...

volumio:
Sorry but I noticed that the modules can be compiled. Could you send the compiled modules to me? in a tar from / ?
So I can add them now!

macmpi:
any hint on the potential kernel version you are contemplating? (may help check some stability issues)

volumio:
No, I mean for the current build. Basically doing what you suggest here:
http://blokas.io/pisound/docs/software/

macmpi:
@volumio my question was for generic purpose, not specific to that issue. Sorry for hijack.

gtrainavicius:
@volumio Thank you, I will make a package for you in the nearest future!

volumio:
Thanks!

gtrainavicius:
@volumio Ok, I've uploaded a build of the module here: http://blokas.io/pisound/pisound_volumio.zip as well as fixed the conflict in dacs.json.

pisound.dtbo should go to /boot/overlays/
4.4.9-v7+/snd_soc_pisound.ko should go to /lib/modules/4.4.9-v7+/kernel/sound/soc/bcm
4.4.9+/snd_soc_pisound.ko should go to /lib/modules/4.4.9+/kernel/sound/soc/bcm

I have tested the build using RPi3 by first building it on a fresh install of volumio-2.118-2017-03-07-pi.img, copying the files over to my laptop, then I freshly installed the same image again, putting the files in the appropriate places, modifying dacs.json, and it worked correctly.

p.s. I had to struggle a bit with 'volumio kernelsource' - the /boot/.firmware_revision contained a revision for the 4.9.x kernel branch, so I had to force it to the one matching 4.4.9 version that is used by Volumio:

`echo 4ce3f4a00648faaa73c7e7d41ff09be83a25bc0a > /boot/.firmware_revision`

gtrainavicius:
Also, is it possible to force the 'mixer type' to software from dacs.json? It'd be useful - our card doesn't expose hardware controls for volume, so volume control must happen in software.

volumio:
No, unfortunately its not possible. Software mixer must be enabled manually, since it degrades SQ. I will send you an image today with the support for PiSound.
Will the input work?

volumio:
Please next time send them with this format:
http://repo.volumio.org/Volumio2/Firmwares/pisound_volumio_4.4.9.tar.gz

Also, before integrating we need to make sure that they are GPL and we're GPL compiant:

Please confirm that:

- The code submitted is released under GPL (indicate  the version) 
- Please show where sources for the binaries can be found
- You will take all responsabilities and obligations concerned by the distribution of the binaries you provided. This includes making the sources available to everyone requiring them. 
- You agree to free Volumio and INTUITU di Michelangelo Guarise from any responsability deriving from distribution of the binaries you provided, and to provide eventual indemnifycation for damages

gtrainavicius:
About the mixer - we will have to add note about it in our documentation.

The code is licensed using GPLv2:

- snd_soc_pisound.ko: https://github.com/raspberrypi/linux/blob/rpi-4.9.y/sound/soc/bcm/pisound.c
- pisound.dtbo: https://github.com/raspberrypi/linux/blob/rpi-4.9.y/arch/arm/boot/dts/overlays/pisound-overlay.dts

We agree to the terms.

gtrainavicius:
@volumio I just tried it out, the pisound.dtbo appears to be in /boot/, while it should be placed in /boot/overlays/ instead.

After manually fixing pisound.dtbo location, it works and without a need to restart.

p.s. We'd prefer the name to appear in lower case letters only, 'pisound'. :)

volumio:
Ok, then this should be the final one.
http://updates.volumio.org/pi/volumio/2.126/volumio-2.126-2017-03-20-pi.img.zip

Let me know how it goes, if everything is allright I'll commit the changes

gtrainavicius:
It works successfully! Thank you!

PR #976, opened 2017-01-24
Fixes Repeat mode in the State Machine module

Callenberg:
This fix takes care of the case when Repeat mode is on by adding two new conditional statements in the method CoreStateMachine.prototype.increasePlaybackTimer:
1.) It checks if the current track is the last track of the queue then the next track is set to the first track in the prefetch phase, unless Consume mode is on.
2.) At the end of the current track, if that track is the last track it sets the current track to the first track unless Consume mode is on.

If the consume mode also is on the last track has to be consumed. No need for next track.

It has been tested also with the other modes. Remember that Random and Consume override Repeat in MPD.
Repeat + Random = Random;  
Repeat + Consume = Consume; 
Random + Repeat + Consume should be Repeat + Consume, but this combination is still buggy (same bugs as before).

### DIFF
--- a/app/plugins/system_controller/i2s_dacs/index.js
+++ b/app/plugins/system_controller/i2s_dacs/index.js
@@ -437,7 +437,7 @@ ControllerI2s.prototype.enableI2SDAC = function (data) {
 ControllerI2s.prototype.writeI2SDAC = function (data) {
 	var self = this;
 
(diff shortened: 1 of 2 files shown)


volumio:
Thanks! I'm testing it!

fanciulli:
Hello @Callenberg 
it would be great if you can make your code easier before merging. lines like the following:

if(this.currentRepeat && ((this.playQueue.arrayQueue.length-1)==this.currentPosition)!== this.currentConsume)

may be complex to debug in the future.

Callenberg:
Okay, I defined a variable for the expression: (this.playQueue.arrayQueue.length-1)==this.currentPosition called isLastTrack and the conditional statements now look better and it might be easier to debug, hopefully.
What do you think?
(Another way is to branch the conditional statements, I don't know what is best)

fanciulli:
Ok for me

PR #978, opened 2017-01-26
group album list by albumartist instead of artist

rezeptpflichtig:
it's the first time I am working with github D: so I hope I am doing this right.


volumio:
Thanks!

PR #986, opened 2017-02-01
UPDATED Repeat + Random + Consume + Repeat Single modes

Callenberg:
This patch hopefully solves the conflict with #995 RepeatSingle.

It fixes the following compared with the master:
Default play mode plays all tracks in the queue sequentially, before it (sometimes) got stuck on one track.
The combination Consume and Random plays tracks randomly instead of sequentially, the same with the combination Random, Repeat and Consume.
Occasionally "ghost track" could be played before, no more playing of tracks that are not in the queue.
Occasionally the wrong track was indicated in the Queue tab, now the current track is indicated correctly.

_Single modes_:
**Repeat** will play the playlist in sequential order indefinitely.
**Random** will play the playlist in a random order indefinitely. 
**Consume** removes each track from the playlist after it has been played. The playlist is sequentially played and stops when the playlist is empty.
**Repeat Single** plays the current track indefinitely.

_Combinations_:
**Random** + **Repeat**  works like **Random**.
**Random** + **Consume**  means that tracks are played randomly and then consumed.
**Repeat** + **Consume**  will eventually consume the whole playlist independently of where the start track was in the list.
**Random** + **Repeat** + **Consume** works like **Random** + **Consume**.
**Single** + **Random** works like **Single**.
**Single** + **Repeat** works like **Single**.
**Single** + **Consume** plays the track and then it is removed from the list.
**Single** + **Random** + **Repeat** + **Consume** as above.

Single requires that Repeat is on in order to work. Defined by #995 RepeatSingle.


chsims1:
Well Callenberg if this all works fine then I think it is a great contribution ... ty in advance for something that has been annoying me for a while. I have to say that I am a little confused by your "combinations."  Why, for example is "Random + Repeat + Consume works like Random + Consume" just not 'Random + Consume?'  Probably showing my ignorance here :)

Callenberg:
As you know it is possible to click and activate more than one play mode at the time. You can even click on all three play mode buttons and have the combination "Random + Repeat + Consume". 

In MPD the combination of Shuffle + Repeat means Repeat only. The good people that implemented MPD thought that shuffle have to override repeat since they are contradictory modes so I implemented that as well,  random + repeat = random.  Repeat is contradictory to random as well. You have to choose one of them.

Shuffle is not exactly the same as random but very very similar... one important difference however is that the implementation of random in Volumio plays the queue indefinitely, shuffle doesn't.

If the play modes random, repeat and consume are combined I let random override repeat and the combination effect has then to be repeat + consume.

Callenberg:
I am afraid I didn't explain what I meant that well, tend to mix up the user experience with implementation details.

This is what I mean: as a user you will not see any difference if you have random and repeat mode on at the same time or just random mode. The list is played indefinitely. The list is in some sense repeated. (random + repeat = random)

If you as a user in addition also turn on the consume mode you will notice a difference since the tracks played will be consumed. (random + repeat + consume = random + consume).

=====
From a Volumio specific implementation point of view there are some cases that must to be dealt with:

If repeat mode is on and the last track is played then the next track has to have index 0 (start at the beginning), except if random mode is also on then let the random function decide next track.

If random mode is on it is better to let the random function not pick the same track two times in a row. The reason is that the next track is set in the prefetch phase and if consume mode is on as well there is a chance that consume will delete the next track in the next phase when the index is actually set. 

If both consume and random mode is on the index of the next track has to be adjusted if the deleted track was before the next one in the playlist, otherwise the prefetch will not work as intended.

Callenberg:
Solved the conflict with the new play mode Repeat Single.

volumio:
We were analyzing the patch, since the previous one resulted in a regression... However, I confirm everything works apart consume!

Callenberg:
Yes, consume is tricky, what is going wrong?

Callenberg:
My reasoning for closing it was that I will never keep up with the frequent changes that is done. The very small code segment I was patching was changed two times after I requested the merge.

But tell me why consume is faulty, may be I can fix that...

volumio:
Yes sorry for that, but we needed to fix this regression quickly. 
Basically consume does not removed played songs once they are played

PR #1001, opened 2017-02-10
Added pHAT DAC, Speaker pHAT and Picade HAT

Gadgetoid:
This change adds current Pimoroni DAC products: pHAT DAC, Speaker pHAT and Picade HAT.

All three use the HiFiBerry DAC overlay, however Speaker pHAT will require some further work to enable support for the built-in VU Meter.


volumio:
Thanks! I'll test and merge!

PR #1006, opened 2017-02-14
Add soundcloud & tunein support

ppmingot:
Hello,

As volumio is one of the most updated and active raspberry Pi HIFI distro, it a must to have soundcloud and tune-in built-in support.
These 2 plugins are already supported by default by mopidy, so you can have these 2 features running out of the box with low effort from your side.

Mopidy also support a lot of hot features, to retrieve the complete list of plugins supported by mopidy : 
_sudo pip search mopidy_


volumio:
Hi, sorry but what is this pull request about?
As for other streaming service, mopidy sounds less good than mpd, and since we care mostly about sound quality we will not change to mopidy. Don't worry, as we will enable the most streaming plugins possible...

ppmingot:
The goal id simply to have soundcloud & tunein supported in volumio.
I share mopidy because it's way to archive.
as you must known mopidy is based on MPD

volumio:
Mopidy is not based on mpd, it just uses the same control protocol

ppmingot:
Ok.
Can you just answer the original question :
Do you plan to support soundcloud, tunein?

volumio:
Yes, we are planning to support as much streaming services as possible.

PR #1008, opened 2017-02-15
Partial rewrite of volumecontrol.js

Gadgetoid:
This is an ongoing partial rewrite of volumecontrol.js to try and reduce code duplication and make it more explicit about what's expected of the `alsavolume` method.

It's a *work in progress*, and this PR is to increase visibility on that work and invite suggestions, comments and criticism- I don't expect it to be merged!

I would like to eliminate some of the global state contained within this code, and perhaps we could add support for getting config params explicitly as numbers, with a default value fallback. For example:

```
volumesteps = this.commandRouter.executeOnPlugin('audio_interface', 'alsa_controller', 'getConfigParamAsNumber', 'volumesteps',10);
```

I'd like, also, to expand this plugin with the ability to periodically sync volume from alsamixer to the front-end- to stay in step with any processes that might adjust the volume behind the scenes.


volumio:
Very very well done. Doing some tests before merge

volumio:
Great work, its much faster now, exactly what we needed. There are a couple of things to fix, but very encouraging work, thanks

Gadgetoid:
You're welcome- it was a rather tentative PR since I'm not exactly a NodeJS wizard, but I thought it could use some tweaks. Thanks for the positive feedback!

PR #1017, opened 2017-02-18
Make retrieval of share size a little more robust.

xipmix:
Quote the NAS mountpoint when calling 'df' and return just the wanted value
with no units, so computing a displayed size becomes simpler.

Fixes: #856


volumio:
Excellent, so much faster!

Issue #627, opened 2016-08-04
Volumio RC2/ Odroid C2: Choosing the NAS from the browse feature (latest version directly built from repo)


gkkpch:
Noticed a problem with adding new music the My Music, the latest version seems to have a problem with selecting music libraries from detected NAS devices. It fills/displays the alias as the name of the selected music directory, as NAS IP address it takes the alias of the NAS device and path is correctly shown as the name of the selected music directory. As this is saved as config parameters it might explain why people have problems when connecting to their NAS after rebooting. 

When I enter all values manually, it seems to work ok.

xipmix:
I took a look at this. The story starts in the Volumio2-UI code, where src/app/plugin/core-plugin/network-drives-plugin.controller.js does this when you pick a share from the detected list:
```
selectShare(share, disk) {
    this.drive.ip = share.name;
    this.drive.name = disk.sharename;
    this.drive.path = disk.path;
}
```
The input to this comes from discoverShares which calls ```smbtree``` to do the discovery and uses the results to create a json data structure, eg:
```
{ nas:
  [  { name: "littlebox", shares: [ { sharename:"A share", path:"A share" } ] },
     { name: "bigbox", shares: [ { sharename:"share1", path:"share1" },
                                               { sharename:"share2", path:"share2" }
                                             ]
     }
  ]
}
```
Let's relate that to your observations, assuming you pick ```\\littlebox\A share```:
 * fills the alias as the name of the selected music directory.
   So you should get 'A share'. That's how it works for me (2.041).

 * as NAS IP address it takes the alias of the NAS device.
   You should get 'littlebox' here. This is the netbios name the NAS sends out in response to browse queries. To connect up the share, volumio uses ```mount.cifs``` and _if netbios name resolution is working_, the mount does actually happen. It might be more robust to have discoverShares do an nmblookup and if it works, add the returned IP address to the data structure. Because IP addresses might change over time (dhcp pools), the name resolution would have to be checked before every mount attempt.

 * path is correctly shown as the name of the selected music directory.
   As you note 'sharename' and 'path' are the same thing. This is deliberate; discoverShares puts the exact same string in both elements of the structure.

So I think the behaviour you describe is what is intended. Why are there problems with reconnecting ? Probably related to the samba config in volumio, but also likely to depend on what's going on in the network around it. If you can recall specific cases it might help move things along.

Question: when you enter all values manually, I assume you use an IP address in the IP address field, not a netbios name?

PR #1029, opened 2017-02-24
Improve discoverShares()

xipmix:
I started looking into #627 and noticed that some shares were not being detected properly by discoverShares(). I've fixed that and hopefully improved the readability somewhat. This has been tested on 2.041 and seems to work correctly.

### DIFF
--- a/app/plugins/system_controller/networkfs/index.js
+++ b/app/plugins/system_controller/networkfs/index.js
@@ -729,7 +729,7 @@ ControllerNetworkfs.prototype.discoverShares = function () {
 	var defer = libQ.defer();
 	var sharesjson = {"nas":[]};
 	try {
-		var shares = execSync("/bin/echo volumio | /usr/bin/smbtree", { uid: 1000, gid: 1000, encoding: 'utf8', timeout: 10000 });
+		var shares = execSync("/usr/bin/smbtree -N", { uid: 1000, gid: 1000, encoding: 'utf8', timeout: 10000 });
 	} catch (err) {
 		var shares = err.stdout;
 	}
--- a/app/plugins/system_controller/networkfs/index.js
+++ b/app/plugins/system_controller/networkfs/index.js
@@ -741,13 +741,13 @@ ControllerNetworkfs.prototype.discoverShares = function () {
 
 	    <domain>
 	    \t\\<nas1>   \t\t<description>
-	    \t\t\\<nas1>\<share1>\t<description1>
-	    \t\t\\<nas1>\<share2>\t<description2>
+	    \t\t\\<nas1>\<share1> \t<description1>
+	    \t\t\\<nas1>\<share2> \t<description2>
 	    \t\\<nas2>    \t\t<description>
 	    \t\t\\<nas2>\IPC$	  \t<description>
 
-        The NAS name field is padded with trailing spaces,
-        as are hidden shares. Non-hidden shares are not.
+        Field names shorter than 16 characters are padded with trailing
+        spaces, to 16 characters. Longer names are not padded.
 	 */
 	try {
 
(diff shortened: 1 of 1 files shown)


volumio:
Pretty clever! (the -N, didn't notice that)
However I am not able to find any share now...

xipmix:
Oh dear. Does the smbtree call work on the command line? If you can send the output I can take a look.

volumio:
Yep, it works:

volumio@volumio:~$ /usr/bin/smbtree -N -b
WORKGROUP
        \\VOLUMIO                       Volumio Audiophile Music Player
                \\VOLUMIO\IPC$                  IPC Service (Volumio Audiophile Music Player)
                \\VOLUMIO\Internal Storage      Volumio Internal Music Folder
        \\FRITZ-NAS                     FRITZ!Box
                \\FRITZ-NAS\IPC$                IPC Service (FRITZ!Box)
                \\FRITZ-NAS\FRITZ.NAS
        \\DISKSTATION
                \\DISKSTATION\IPC$              IPC Service ()
                \\DISKSTATION\video             System default shared folder
                \\DISKSTATION\Torrent
                \\DISKSTATION\Share con spazi
                \\DISKSTATION\Programmi
                \\DISKSTATION\photo             System default shared folder
                \\DISKSTATION\NetBackup         System default shared folder
                \\DISKSTATION\music             System default shared folder
                \\DISKSTATION\Mi
                \\DISKSTATION\IMGs
                \\DISKSTATION\homes             user home
                \\DISKSTATION\FLAC
                \\DISKSTATION\Film
                \\DISKSTATION\Cose Mondo
                \\DISKSTATION\Backup
        \\DESKTOP-HSL3GOL


So I think its a parsing error

balbuze:
Even with 2.0.96 my nas is not always detected. Sometime it's ok, sometime is not. And /usr/bin/smbtree -N -b return nothing in this case, even if nas is already mounted.

xipmix:
Somehow I typoed 'excludednas' as 'excludenas'. This caused an undefined variable exception but I guess since it was inside a try/catch, we ended up in the catch. I've changed the variable name to make it harder to get wrong. For my penance I cut and pasted the updated code into a separate file and ran test cases with your smbtree output and mine, to confirm the parsing worked correctly.

volumio:
Does not work... Now the scan goes OK, but when I hit save it reports "no such address"

xipmix:
Sorry about that. I did some more testing. The shares I had been using to test with happened to have names longer than 16 characters, except the hidden shares. The long names are not padded, but any share names shorter than 16 characters get padded with trailing spaces to be 16 chars long. I wrongly assumed that only hidden shares (which _were_ short names) were padded. I didn't pick this up from your pasted smbtree output because the tabs were lost by the markdown rendering.

So why the failure? The share name gets used as the alias, which in turn is used to construct the mountpoint. The mountpoint is ok because we swap spaces with underscores. For the name of the remote share, the spaces are correctly escaped in addShare but the spaces are not actually part of the share name as defined in smb.conf on the remote host. So the mount fails because we are asking for a share name that does not exist.

Example log:
```
sudo[5374]: volumio : TTY=unknown ; PWD=/ ; USER=root ; COMMAND=/bin/mount -t cifs -o guest,ro,dir_mode=0777,file_mode=0666,iocharset=utf8,noauto //UBUNTU/var-tmp         /mnt/NAS/var-tmp________
...
volumio[4939]: exec error Error: Command failed: /usr/bin/sudo /bin/mount -t cifs -o guest,ro,dir_mode=0777,file_mode=0666,iocharset=utf8,noauto //BIGBOX/var-tmp\ \ \ \ \ \ \ \  /mnt/NAS/var-tmp________
volumio[4939]: Retrying with upper case share name
volumio[4939]: mount error(6): No such device or address
volumio[4939]: Refer to the mount.cifs(8) manual page (e.g. man mount.cifs)
volumio[4939]: info: Error mounting  var-tmp         on IP UBUNTU : (6): No such device or address
volumio[4939]: Refer to the mount.cifs(8) manual page (e.g. man mount.cifs)
```
So if we strip the trailing spaces things should work. The number of users with trailing spaces in the real share name is probably zero.

volumio:
Yes that works! And its also quite faster to mount, well done!

PR #1038, opened 2017-03-01
Bugfix for addShare

xipmix:
Apologies - I did not check the uuid already defined case...


xipmix:
On reflection this probably fixes #1047 (no smb mounts after OTA upgrade)

PR #1050, opened 2017-03-06
Clean up submitted logs somewhat

xipmix:


volumio:
Good one, I'll test and merge if OK

PR #1053, opened 2017-03-06
Fixes issue

volumio:


Callenberg:
Hello,
I think there should be only one conditional statement on line 332, check if Random-mode is true. There is no need to involve the state of the Consume-mode anymore. As now Repeat will only work correctly  if Consume-mode is also true.

The reason that it originally was a  ` !== this.currentConsume` in the conditional statement for Repeat-mode was not to interfere with Consume-mode at the time. I was guilty of that.

. . .  but I haven't had time to check this fix yet, will do. I just saw the diff.

Kind regards,

PR #1056, opened 2017-03-07
Rewrite of setting of Random, Repeat and Single Repeat play modes + Consume

Callenberg:
Here is a proposal for the code that sets the actual play modes (Random, Repeat and Single Repeat, and Consume (see below). 
If the play modes are combined it works like this:
Random + Repeat = Random
Random + Single Repeat = Single Repeat
Repeat + Single Repeat = Single Repeat
Random + Repeat + Single Repeat = Single Repeat

Consume is mentioned below.

### DIFF
--- a/app/statemachine.js
+++ b/app/statemachine.js
@@ -318,7 +318,7 @@ CoreStateMachine.prototype.increasePlaybackTimer = function () {
 			this.askedForPrefetch=true;
 
 			var trackBlock = this.getTrack(this.currentPosition);
-                        var isLastTrack = (this.playQueue.arrayQueue.length - 1) == this.currentPosition;
+      var isLastTrack = (this.playQueue.arrayQueue.length - 1) == this.currentPosition;
 
 			this.nextIndex = this.currentPosition + 1;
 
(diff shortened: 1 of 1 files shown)


Callenberg:
In addition a possible solution for Consume play mode. 

If Consume is combined with the other modes it should work like this:
Random + Consume = Random and each track is also deleted from the queue when it is finished 
Repeat + Consume = Repeat and each track is also deleted after it is finished
Repeat Single + Consume = plays the track once and then the track is deleted, the playback stops
Random + Repeat + Consume = Random + Consume
Random + Repeat + Repeat Single + Consume = Consume + Repeat Single

volumio:
Sorry, we can't merge this as we're in the middle of a substantial rewrite of this. Thanks anyway and sorry again

Callenberg:
No problem, I think it is great that you are fixing these things. It doesn't have to be my code, the important thing is that it will eventually work. Great!

 Thanks


________________________________
Från: Volumio <notifications@github.com>
Skickat: den 13 mars 2017 16:37
Till: volumio/Volumio2
Kopia: Callenberg; Author
Ämne: Re: [volumio/Volumio2] Rewrite of setting of Random, Repeat and Single Repeat play modes + Consume (#1056)


Sorry, we can't merge this as we're in the middle of a substantial rewrite of this. Thanks anyway and sorry again

PR #1057, opened 2017-03-08
More robust mounting

xipmix:
After tweaking the mountpoint paths a while ago, I started looking at strange paths on the remote share. After trying all sorts of quoting & escaping schemes I realised the problem was in linux-mountutils.

I've tested this with a few different strangely-named samba shares and tested mountpoints with things like () and ' in them, for CIFS and NFS. More testing would be welcome. Existing shares _should_ continue to work, as these changes don't affect how the mountpoint string is constructed.

The patch to linux-mountutils has been sent upstream but might have to be carried for a while.

### DIFF
--- a/app/plugins/system_controller/networkfs/index.js
+++ b/app/plugins/system_controller/networkfs/index.js
@@ -160,10 +160,10 @@ ControllerNetworkfs.prototype.mountShare = function (data) {
 
 	var fstype = config.get('NasMounts.' + shareid + '.fstype');
 	var options = config.get('NasMounts.' + shareid + '.options');
-	var pathraw = config.get('NasMounts.' + shareid + '.path');
+	var path = config.get('NasMounts.' + shareid + '.path');
 	var mountidraw = config.get('NasMounts.' + shareid + '.name');
 	// Check we have sane data - operating on undefined values will crash us
-	if (fstype === 'undefined' || pathraw === 'undefined') {
+	if (fstype === 'undefined' || path === 'undefined') {
 		console.log('Unable to retrieve config for share '  + shareid + ', returning early');
 		return defer.promise;
 	}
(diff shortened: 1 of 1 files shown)


volumio:
Fails to mount:
info: Error mounting  Ultra Hi Res Music on IP NAS02E135 : exec error Error: Command failed: /usr/bin/sudo /bin/mount -t cifs -o guest,ro,dir_mode=0777,file_mode=0666,iocharset=utf8,noauto //NAS02E135/Ultra Hi Res Music /mnt/NAS/Ultra_Hi_Res_Music

xipmix:
Thanks I will do some more testing.
I assume this was with linux-mountutils patched?

volumio:
MMM no.. Which patch?

xipmix:
> MMM no.. Which patch?
 
In the changelog of the first patch.
See also https://github.com/mw-white/node-linux-mountutils/pull/1

volumio:
Interesting, for the time being we could include the code as a library until it gets merged. pretty much like I did with wireless on network. What do you think

xipmix:
That's exactly the response I was hoping for!

--

volumio:
Great! If you update the PR I'll test it and eventually merge!

xipmix:
Could you clarify how I would update the PR?
I guess you mean I need to submit a patch for linux-mountutils.
I understand that the build.sh pulls in the node_modules tarball
but how does that tarball get updated?
I seem to be missing something here.

volumio:
Basically, you create a lib folder inside the networkfs folder. Place there the index.js (and related files, if needed) of mountutils there. Then you call this library instead of the node module by calling it instead of 

var mountutil = require('linux-mountutils');

it will be

var mountutil = require('./lib/mountutil.js');

See how I did here:
https://github.com/volumio/Volumio2/tree/master/app/plugins/system_controller/network

mw-white:
This looks great, thanks...I like your test directory names.

xipmix:
Upstream has made 1.0.1 available. Should I wait for volumio/greenkeeper to pick up that version?

volumio:
No, I'll take care of it. Thanks!!

Issue #1031, opened 2017-02-24
Edit DNS via UI


macmpi:
Just a placeholder in reference to [this](https://github.com/volumio/Build/commit/077e503f73504dec0d602a98aa377313cfe3b0f9#commitcomment-20985929), so that some Network Preference Panel UI section is added to allow manual DNS entry: this may greatly help users having name resolution issues in specific cases (ISP filtering, bugs, etc).

macmpi:
Thanks @volumio for [this](https://github.com/volumio/Volumio2/commit/5fed4010f1c5d15df9f98e32cbd95a56707184c4). Closing.

macmpi:
A second thought: would be nice if current DNS values are displayed (like for IP addresses), as it is key to make a decision whether or not DNS are correct and need a change.

macmpi:
Actually I'll try to better express the requirement as a scrum-like user story.
As a user, I want to be able to see which are the actve DNS used by the system, and flip settings between automatic DHCP discovery, and custom manual entry.

macmpi:
@xipmix wrote:

> I think adding the custom nameserver lines to resolv.conf.head is better that adding to the resolv.conf.tail, but it would be nice to be able to opt out of using the google serversas well.

Yes.
Actually I think that with the new option to manually enter DNS info now, we can safely remove `resolv.conf.tail` altogether: this will indeed opt-out from Google servers, unless users positively decides to put them in manually.
`resolv.conf.tail` had been put-in initially as a workaround to fix some DNS issues, but it's a moot point now.

So let's:
- have manual entry as `resolv.conf.head`
- discard `resolv.conf.tail` from [Build script](https://github.com/volumio/Build/blob/63f44cb8f560ec3f64dbacba6536bb74d6756622/scripts/volumioconfig.sh#L432-L435) (note: OTA updated devices will keep it until reflash)
- have UI tweaks to show current DNS settings and allow to flip between auto and manual entry.

I wish I were more fluent in node and js so that I would do the PR myself...

macmpi:
Started to adapt to `resolv.conf.head` in https://github.com/volumio/Volumio2/pull/1061 PR
I realized that there is a critical UI issue currently **which may impact current users having DNS problems**:
- ON/OFF should activate or clear the .tail file (or move in/out with other name), and restart network
- SAVE should should not do anything to the file if OFF

Still need further UI work to display current DNS entries of active `/etc/resolv.conf` file

macmpi:
[...]

PR #1061, opened 2017-03-10
Fixing Manual DNS entry

macmpi:
Use /etc/resolv.conf.head instead of /etc/resolv.conf.tail
As per https://github.com/volumio/Volumio2/issues/1031

Still need to fix:
- UI logic rework (critical)
------ ON/OFF should activate or clear the .head file (or move in/out with other name), and restart network
------ SAVE should should not do anything to the file if OFF
- UI display to show current DNS server used by /etc/resolv.conf

### DIFF
--- a/app/plugins/system_controller/network/index.js
+++ b/app/plugins/system_controller/network/index.js
@@ -844,15 +844,15 @@ ControllerNetwork.prototype.saveDnsSettings = function (data) {
 		config.set('secondary_dns', data.secondary_dns);
 	}
 
-	var dnsfile = '##custom DNS' + os.EOL + 'nameserver '+ data.primary_dns + os.EOL + 'nameserver '+ data.secondary_dns;
-	exec("/usr/bin/sudo /bin/chmod 777 /etc/resolv.conf.tail", {uid: 1000, gid: 1000}, function (error, stdout, stderr) {
+	var dnsfile = '##custom DNS' + os.EOL + 'nameserver '+ data.primary_dns + os.EOL + 'nameserver '+ data.secondary_dns + os.EOL;
+	exec("/usr/bin/sudo /bin/chmod 666 /etc/resolv.conf.head", {uid: 1000, gid: 1000}, function (error, stdout, stderr) {
 		if (error !== null) {
-			console.log('Canot set permissions for /etc/resolv.conf.tail: ' + error);
+			console.log('Canot set permissions for /etc/resolv.conf.head: ' + error);
 		} else {
-			self.logger.info('Permissions for /etc/resolv.conf.tail')
-			fs.writeFile('/etc/resolv.conf.tail', dnsfile, function (err) {
+			self.logger.info('Permissions for /etc/resolv.conf.head')
+			fs.writeFile('/etc/resolv.conf.head', dnsfile, function (err) {
 				if (err) {
-					self.logger.error('Cannot write wpasupplicant.conf ' + error);
+					self.logger.error('Cannot write DNS File' + error);
 				} else {
 					self.commandRouter.wirelessRestart();
 					self.commandRouter.networkRestart();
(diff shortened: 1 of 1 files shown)


macmpi:
Note: do NOT merge until UI logic issue is fixed

volumio:
If you could fix the UI logic that would be really helpful...

volumio:
Thanks, so is it functionally complete?

macmpi:
UI flow is indeed now fixed.
(Note this PR should be merged at same time with [that one in Build repo](https://github.com/volumio/Build/pull/182)).

It now operates as expected, but still an **active DNS feedback** UI is highly desirable, so that user can detect eventual problems and decide to input different values with new settings: it would most probably make sense to have it within Network status sections on top of the page.

It would then likely fit into `ControllerNetwork.prototype.getInfoNetwork` with a new `dnsstatus` of a type/format to be defined and displayed accordingly (not sure how to do that).
It would then report up to the first 2 `nameserver:` lines of `/etc/resolv.conf`

volumio:
I think it should work like this:
- tail with 8.8.8.8 and 8.8.4.4 as fallback 
- head to be written only if custom dns= true
- if if custom dns= true remove tail

We should cover every case with this. What do you think

macmpi:
Thanks for feedback.
Actually I don't think we want to have `.tail` at all now: fallback is not needed anymore now we can set custom DNS: custom DNS is the actual fallback; and it's a fallback user can now fully control.

However, note that default custom DNS (as provided by UIConfig) are indeed Google servers, so just enabling custom DNS with default settings will indeed provide Google DNS: so we are covered with proven fallback.
So, in short I think only `.head` matters now, as with current PR proposal.
Makes sense?

volumio:
Yes, theoretically all you say is correct. However, there were cases where people could not browse webradios or find updates. Because somehow dhcpcd did not catch their router's DNS.

So, for those edge cases I think we still need the tail fallback. One of our goal is simplicity: and for a non expert user this could be difficult to debug.

macmpi:
If `.tail` is set as such, it may indeed allow some people to use internet feature with failing auto DNS config, but then they will also un-noticeably be tracked by Google DNS services without knowing how to opt-out of this. They won't even know they have such potential DNS issue either.
(and BTW a certain number of ISPs are actively blocking Google DNS, which won't be of any help then).

The key missing bit I think, is a **UI feedback of actively used DNS servers** by Volumio: this is probably the most common-way to figure-out problems causes & give clues to users for fixing the issues. That UI area could even report a warning if no server is listed there (i.e. in case of failing auto-DNS, the problem we want to treat), advising to set manual DNS (which default to Google-ones, but can be customized at user-will).
Wouldn't this be more user-friendly and privacy-cautious than potentially setting a back-door to Google services without user consent?

volumio:
What about open DNS instead of google ones as fallback?
For the UI feedback, we need to find a reliable way to understand which DNS servers are in use. Do you have any idea for that?

macmpi:
openDNS (208.67.222.222, 208.67.220.220) might be a less concerning option for some, and may be less filtered by ISPs...It could also be the default in UIconfig BTW.
Unsure they require account subscriptions now. 

For UI feedback, a good measure would be to check `nameserver` entries within `/etc/resolv.conf`, and list the first 2 (if such exist). If none exist, then issue a visual feedback to advise setting DNS manually. Note: is if `.tail` exist, this warning test need to filter away such `.tail `entries as they will always exist. This makes such helpful test more complex (in particular if user does willingly set OpenDNS).

So do you want me to add-back `.tail` with such OpenDNS entries in that PR (and [assorted Build PR](https://github.com/volumio/Build/pull/182))?

volumio:
Yes, thank you!

macmpi:
will try to do, test, and submit tomorrow.
Will you take care of the DNS feedback UI in Network status sections on top of the page?

macmpi:
Updated as suggested to enable OpenDNS defaults.
Should be merged at same time with https://github.com/volumio/Build/pull/182

PR #1065, opened 2017-03-13
Fixing and completing German translation

Jotam:
After updating Volumio from version 2.114 to 2.118, the German translation could not be selected anymore. So I reworked to whole language file to fix this. I also added the remaining translations that had previously not been available in German. The German language file should now be complete. (I used strings_en.json as template.)


volumio:
Thanks!

Jotam:
Any time. Maybe you should publish another point release very soon. In 2.118 the German language settings doesn't work completely. When I tried to switch from English to German, the setting simply wasn't applied. (And I used a fresh install to try this.) I bet this bug will annoy many German speaking users. :-)

volumio:
I am doing this as we speak!

Jotam:
Awesome! Thanks a lot!

PR #1069, opened 2017-03-14
Working with share configuration objects

xipmix:
This is cleanup, no functional changes. The focus is on making the way fields accessed with config.get() is consistent across functions within the module.
Tested on 2.118 + linux-mountutils-1.02 + the 'more robust mounting' patch. Add/edit/delete of shares seemed to work ok.

### DIFF
--- a/app/plugins/system_controller/networkfs/index.js
+++ b/app/plugins/system_controller/networkfs/index.js
@@ -362,10 +362,9 @@ ControllerNetworkfs.prototype.addShare = function (data) {
 	}
 
 	uuid = libUUID.v4();
-	var key = "NasMounts." + uuid + ".";
 	self.logger.info("No correspondence found in configuration for share " + name + " on IP " + ip);
 
-	var saveshare = self.saveShareConf(key, uuid, name, ip, path, fstype, username, password, options);
+	var saveshare = self.saveShareConf('NasMounts', uuid, name, ip, path, fstype, username, password, options);
 
 	saveshare.then(function () {
 		var mountshare = self.mountShare({key:uuid});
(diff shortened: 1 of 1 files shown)


volumio:
What's the improvement of this?

xipmix:
It makes the way keys are constructed and used consistent across
functions in the module, which I think helps with readability/
maintainability.
In mountShare it removes a lot of 'Nasmounts'+shareid+'.thing'.

--

volumio:
Altough I trust you, we need to extensively test it before merge (trying to become a disciplinate project manager ehehe). What about setting up a battery of tests for the NAS function?

xipmix:
I tried to break the changes up so that you would be reassured they
aren't going to break things, but just on general principles it
would be good to have tests to avoid regressions from refactoring.

I have no real idea how to set it up, though, apart from knowing
that 'mocha' exists.
What are you using in other parts of the project?
Do you have a test sub-project somewhere?

volumio:
I just implemented mocha last week, now we just need to write tests eheheh.
I am worried about regression so I think our priority now would be making sure we avoid bugs and regressions. In my free time I will start making an automatic test routine. However we will need a separate pi to act as client for it... 
What we could do would be making some unit tests with mocha on what we can..

xipmix:
Happy to help if I can, but I'll need some guidance on how to test
eg networkfs in isolation; mocking etc is all new to me.

volumio:
I checked it and it works, merging

volumio:
As for testing, I will come up with a strategy soon. For now rule is: every new feature = testing all the features impacted by the change. 
In any case you're doing an excellent contribution, thank you!! You really stepped up the quality of the NAS part!

Issue #847, opened 2016-11-15
Plugin manager: some erratic installs


macmpi:
Hi,

I encountered erratic plugin installs several times, particularly when installing a dropped zip in install UI.
Indeed it seems that sometimes installation may happen twice (the second would fail if the first one was successfully registered).
It may be due to insufficient clean-up of temporary install files upon installation failure.
If some install procedure fails, plugin manager may discover remainings of `/tmp/downloaded_plugin.zip` and try to install over-again ([an exemple here](https://volumio.org/forum/spotify-connect-plugin-old-volumio-login-with-winscp-t4524-200.html#p23117)).

Maybe it would be suitable to remove each temporary file & folders after each step even if not successful (`/tmp/downloaded_plugin.zip`, `/tmp/downloadedPlugin` folder, and `/tmp/plugins` folder) or find a way to avoid those multiple installs.

volumio:
Yes, I would say we should understand why installation restarts. Any idea?

macmpi:
I'm afraid no idea at this time.
However /tmp cleanup would at least avoid consequences, while freeing-up some RAM in the meantime.
Re-entry could still be noticed with journalctl messages, so problem does not go unnoticed.

macmpi:
Isn't the issue with [this](https://github.com/volumio/Volumio2/blob/master/http/index.js#L94), which may be called several times in pages refreshes?

After first transfer to `/tmp/plugins/` and install upon `socket.emit('installPlugin', { url:pluginurl});`, [installPlugin](https://github.com/volumio/Volumio2/blob/master/app/pluginmanager.js#L411) does not delete original file from `/tmp/plugins/`after install.
Therefore any re-entry in `app.route('/plugin-upload')` will produce re-install.

macmpi:
While at it, it would be nice if RAM usage can be optimized by limiting plugin duplication in /tmp: indeed installing some big plugins (like spotifyconnect for arm6) on limited RAM devices (such as rpi0) may lead to fairly low RAM conditions.
Here are some suggestions, but you may have better ideas.

For drag&drop install, can't the initial upload be directly into `/tmp/downloaded_plugin.zip` (omitting temporary `/tmp/plugin/myplugin.zip`), and then just test inside installPlugin() from pluginmanager.js, if that file already exists to differentiate drag&drop or network install (and do wget only in the latter case)?
[...]

PR #1073, opened 2017-03-17
s/defere/defer/g

xipmix:
This may help with issue #847 . It's a bit surprising 'use strict' didn't catch this.

### DIFF
--- a/app/pluginmanager.js
+++ b/app/pluginmanager.js
@@ -679,7 +679,7 @@ PluginManager.prototype.rmDir = function (folder) {
 
 	var defer=libQ.defer();
 	fs.remove('/tmp/downloaded_plugin', function (err) {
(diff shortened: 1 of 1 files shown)


volumio:
Yes, you're right. is a typo

macmpi:
@xipmix good catch: have you indeed checked downloading a .zip plugin if it no-longer installs twice?

xipmix:
Can't right now but I was planning on doing that check. Don't let that stop you trying it; simple enough to patch the file on a running instance & restart.

macmpi:
Had a bit of time to test: this fix unfortunately did not fix the multiple install of manually dropped-in plugins (tested with gpio-buttons.zip, and auto_play.zip for instance).
https://github.com/volumio/Volumio2/issues/847 (and https://github.com/volumio/Volumio2-UI/issues/295) are still here.
Still good catch though...

PR #1083, opened 2017-03-22
Avoid  upload failures on certain filenames

xipmix:
The issue referenced below reports the upload stops part way through. This is because the wget command falls over:
```
Mar 22 11:25:35 volumio volumio[3248]: Uploading: auto_play(1).zip
Mar 22 11:25:35 volumio volumio[3248]: Upload Finished of auto_play(1).zip
Mar 22 11:25:36 volumio volumio[3248]: info: Downloading plugin at http://127.0.0.1:3000/plugin-ser
ve/auto_play(1).zip
Mar 22 11:25:36 volumio volumio[3248]: info: Cannot download file http://127.0.0.1:3000/plugin-serv
e/auto_play(1).zip - Error: Command failed: /usr/bin/wget -O /tmp/downloaded_plugin.zip http://127.
0.0.1:3000/plugin-serve/auto_play(1).zip
Mar 22 11:25:36 volumio volumio[3248]: /bin/sh: 1: Syntax error: "(" unexpected
Mar 22 11:25:36 volumio volumio[3248]: info: Error: Error: Error: Command failed: /usr/bin/wget -O 
/tmp/downloaded_plugin.zip http://127.0.0.1:3000/plugin-serve/auto_play(1).zip
Mar 22 11:25:36 volumio volumio[3248]: /bin/sh: 1: Syntax error: "(" unexpected
```
The changes in this PR should resolve this issue, please test.

Tested against 2.118, using a file named 'auto_play(1).zip'.

Fixes: volumio/Volumio2-UI#299

### DIFF
--- a/app/pluginmanager.js
+++ b/app/pluginmanager.js
@@ -422,7 +422,7 @@ PluginManager.prototype.installPlugin = function (url) {
 	self.pushMessage('installPluginStatus',{'progress': 10, 'message': 'Downloading plugin','title' : modaltitle, 'advancedLog': advancedlog});
 
 
-	exec("/usr/bin/wget -O /tmp/downloaded_plugin.zip "+url, function (error, stdout, stderr) {
+	exec("/usr/bin/wget -O /tmp/downloaded_plugin.zip '" + url + "'", function (error, stdout, stderr) {
 
 		if (error !== null) {
 			currentMessage = "Cannot download file "+url+ ' - ' + error;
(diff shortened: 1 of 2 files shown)


macmpi:
Thanks for looking at this indeed: Mac users will certainly appreciate!
Can't we use a fixed name rather going through UUID addition?
How about downloading directly as `downloaded_plugin.zip` and let install as if it was a normal internet plugin from there, as [explained here](https://github.com/volumio/Volumio2/issues/847#issuecomment-261084732)?

xipmix:
The idea of a uuid was to avoid upload collisions,
ie more than one upload in progress at once.
But perhaps that is not necessary.

xipmix:
One other comment - I preferred to keep this fix focused on fixing the problem reported. The other changes  you propose seem like a good idea but can be addressed separately.

volumio:
Clever and elegant, I liked it!

macmpi:
@xipmix 
Would it make sense to apply similar kind of change to background files downloads, which could potentially suffer similar issues if filenames have special characters?

While at it, would also be nice to prefix such user-uploaded background with something (like `user_` for instance), so that files can be conveniently filtered in a back-up/restore procedure...
Displayed background name may omit that prefix though.

Issue #1104, opened 2017-03-29
Reduce duplication of plugins during upload process


xipmix:
In issue #847 @macmpi raised the following issue, which I am posting as a separate one:

While at it, it would be nice if RAM usage can be optimized by limiting plugin duplication in /tmp: indeed installing some big plugins (like spotifyconnect for arm6) on limited RAM devices (such as rpi0) may lead to fairly low RAM conditions.
Here are some suggestions, but you may have better ideas.

For drag&drop install, can't the initial upload be directly into /tmp/downloaded_plugin.zip (omitting temporary /tmp/plugin/myplugin.zip), and then just test inside installPlugin() from pluginmanager.js, if that file already exists (or empty URL passed) to differentiate drag&drop or network install (and do wget only in the latter case)?
Then, can /tmp/downloaded_plugin.zip extraction happen directly into /data/plugins/tmp_plugin (and delete /tmp/downloaded_plugin.zip right-away), and finally move tmp_plugin into final destination with final name once package.json is actually read (and no error found)?

Those (or similar) measures could significantly save RAM in some cases, and by cleaning after each step, would also be safer, while the re-entry issue is being investigated in drag&drop situations.
Thanks for consideration.

PR #1105, opened 2017-03-29
Rework plugin upload (RFC)

xipmix:
This is a Request For Comment more than a real pull request, because I'm proposing a change to the plugin installation API. Instead of re-uploading to the express server on port 3000 it just works at the filesystem level, which is a bit of a shift in paradigm. 

If this is considered a reasonable approach it could also be applied to upload of backgrounds.

It addresses the issue raised in #1104.

Tested on 2.118 with the patches from PR #1083 applied, for both the normal case and the case where the input 'data' object does not have the expected 'sourcefile' field. It seems to me it should be possible to write a mocha test for this but I'm not quite sure where to start.

### DIFF
--- a/app/pluginmanager.js
+++ b/app/pluginmanager.js
@@ -413,7 +413,7 @@ PluginManager.prototype.installPlugin = function (data) {
 	var defer=libQ.defer();
 	var modaltitle= 'Installing Plugin';
 	var advancedlog = '';
-    var sourcefile = data.sourcefile;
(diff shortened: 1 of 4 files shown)


macmpi:
Interesting, thanks.
Can't comment on overall concept change though.
By chance (I know this is not direct motivation for this, and this should probably not be judged by this), does it also happen to fix the the duplicated plugin installs?

xipmix:
No impact on the duplicate plugin installs.

PR #1108, opened 2017-03-29
Fix Artist artwork

liberatorcnx:
Fix Artist artwork and remove redundant code


liberatorcnx:
Please note: We should also include the latest merge from [album-art](https://github.com/lacymorrow/album-art) to solve the problem of having '&' (ampersand) characters in artist names.

volumio:
Great! It works brilliant!

Issue #1070, opened 2017-03-16
Add defensive for volumioPushState and volumioPushQueue method


JedS:
**CoreCommandRouter.prototype.volumioPushState** and **CoreCommandRouter.prototype.volumioPushQueue** call respectively **pushState** and **pushQueue** methods on every plugin belonging to the **user_interface** category.
If a plugin developer forget to implement these methods, the web UI is not updated anymore correctly.
The Plugin API documentation should be updated also.

PR #1118, opened 2017-04-05
Add defensive for volumioPushState and volumioPushQueue method

JedS:
Add extra verification of pushState and pushQueue method on plugin #1070


volumio:
You're right, clever

PR #1127, opened 2017-04-08
Pi Models detection compatibility

macmpi:
Preparation for newer kernel support while ensuring full backward compatibility.
**cpuinfo Hardware** returned chipset families (BCM2708/09/10) before kernel 4.8.y, and returns SoC reference (BCM2835/6/7) since then.


volumio:
Well done!

PR #1134, opened 2017-04-12
Fix Search Function

liberatorcnx:
Fix Search Function


volumio:
It just works great, thank you very much!

PR #1136, opened 2017-04-13
add pre-end-of-life actions mechanism on plugins

JedS:
add 2 methods **onVolumioReboot** and **onVolumioShutdown** for **pluginManager**. These methods call respectively **onVolumioReboot** and **onVolumioShutdown** on every enabled plugins.
Update the CoreCommandRouter reboot and shutdown methods, calling the pluginManager.onVolumioReboot  or pluginManager.onVolumioShutdown just before the real reboot or shutdown


volumio:
Well done!

Issue #1114, opened 2017-03-31
Plugin manager: improve load order priority management


macmpi:
Plugins may have a set boot priority order number in `package.json`.
They are then loaded by such boot priority order by [that code](https://github.com/volumio/Volumio2/blob/master/app/pluginmanager.js#L188-L192): loading is essentially about invoking each `pluginInstance.onVolumioStart()` initialization routine.

However current implementation does NOT guaranties plugin(s) with **boot priority 2** will be initialized **after** all plugin(s) with **boot priority 1** initialization is **completed**.
Therefore, this may lead to unpredictable results if a **boot priority 2** plugin depends on higher priority **boot priority 1** plugin initialization being **completed**.
I took examples of 2 adjacent priorities here, but this could actually span over several priority numbers.

It is expected that **onVolumioStart()** should not take too much time to complete (good practice recommendation probably), but nothing can guaranty actual timing: it is at plugin developer's discretion to ensure.

Therefore, it would be very valuable if `pluginmanager` could ensure that plugins **do complete their initialization by boot priority number batches**, at least: all the # 1 start & complete, then all # 2 start & complete, then...

I'm not familiar enough with promises programming to deals with such type of launch queues concept, but I hope this will be considered.

Thanks for feedback.

macmpi:
While trying to manage promises chains in plugin load (onVolumioStart), it seems some plugins have issues:
- 1 _appereance_ and _myvolumio_ plugins miss `return libQ.resolve();` in `onVolumioStart()`
- 2 acces to `PlayerName` seem to cause problems to _upnp_ & _volumio_discovery_ plugins
- 3 volumecontrol.js fails line 37 (if ((cards[device] != undefined) && (cards[device].name != undefined))) 
-...

first item is easy to fix, unsure about others...

macmpi:
@volumio would you have some insight on following log, which looks a bit unexpected:

```
Apr 18 16:34:25 volumio volumio[1116]: info: Loading plugin "system"...
Apr 18 16:34:25 volumio volumio[1116]: info: Loading plugin "appearance"...
Apr 18 16:34:27 volumio volumio[1116]: info: Loading plugin "network"...
[...]
Apr 18 16:34:39 volumio volumio[1116]: info: Loading plugin "rest_api"...
Apr 18 16:34:39 volumio volumio[1116]: info: Loading plugin "websocket"...
Apr 18 16:34:39 volumio volumio[1116]: info: Plugin volspotconnect2 is not enabled
[...]

PR #1145, opened 2017-04-17
pluginmanager: deal with onVolumioStart() promises

macmpi:
plugins' onVolumioStart() promises does not seem to be handled: trying to fix that.

This change reveals potential issue with some plugins then failing to load (missing `return libQ.resolve();`): have been fixed in separate PR.

Please refer to issue https://github.com/volumio/Volumio2/issues/1114


macmpi:
Replaced by https://github.com/volumio/Volumio2/pull/1162

PR #1155, opened 2017-04-29
System plugin: fix onVolumioStart promise

macmpi:
Ensure `deviceDetect()` and `callHome()` are properly managed inside `onVolumioStart()` promise.


macmpi:
We could have a fail-safe in callHome then.

More fundamentally, are deviceDetect() and callHome() really mandatory within onVolumioStart()? (are their result used by other plugins init or start...I do not think so).
Can't they just happen in an onStart() sequence then: this will allow onVolumioStart() to finish efficiently during init phase, and not delay other plugins start procedures when sequences are respected at some point?
What do you think?

PR #1162, opened 2017-05-03
Pluginmanager: improve load, start, stop promise management

macmpi:
This PR just focuses on improving promise management in load/start/stop plugins operations.
It fixes the infrastructure, so that later PR may further enhance load-order priorities management, make startup-sequence more deterministic, etc...
It does benefit right-now to add-on plugin Start/Stop operations which are more reliable.

Some important note:
- all plugins `onVolumioStart()` must now return promises and will crash if not.
Built-in plugins now comply with that, but some 3rd Party plugins may not. I contacted most plugins developers and submitted relevant PRs in [volumio-plugins](https://github.com/volumio/volumio-plugins/pulls): it should be ok when eventually merged. We may want to add a note into plugin [development docs about `onVolumioStart()`](https://volumio.github.io/docs/Plugin_System/Index.js.html) required promises for future reference.
- this PR does not require any other change: current `/volumio/app/index.js` startup sequence works as previously with it and is not affected (nor is fixed then). Startup sequence scheduling fixes will need a separate PR.
- In my testing process, I noticed that `onVolumioStart()` **system plugin** has `deviceDetect()` & `callHome()` calls "leaking" outside of the promise dependency: https://github.com/volumio/Volumio2/pull/1155 intends to fix that.
- similarly `onVolumioStart()` from **i2s_dacs plugin** has `i2sDetect()` call "leaking" outside of the promise, and may need a fix too.

@volumio and @fanciulli I welcome any feedback.

PS: this PR replaces https://github.com/volumio/Volumio2/pull/1145

### DIFF
--- a/app/pluginmanager.js
+++ b/app/pluginmanager.js
@@ -134,11 +134,10 @@ PluginManager.prototype.loadPlugin = function (folder) {
 				// Handle non-compliant onVolumioStart(): push an error message and disable plugin
 				self.coreCommand.pushToastMessage('error',name + " Plugin","This plugin has failing init routine. Please install updated version, or contact plugin developper");
 				self.logger.error("Plugin " + name + " does not return adequate promise from onVolumioStart: please update!");
-				self.disablePlugin (category,name);
 				myPromise = libQ.resolve();  // passing a fake promise to avoid crashes in new promise management
 			}
-			else
-				self.plugins.set(key, pluginData);  // not set if no promise, so can't be started/stopped
+			
+			self.plugins.set(key, pluginData);  // set in any case, so it can be started/stopped
 		
 			defer.resolve();
 			return myPromise;
@@ -262,7 +261,6 @@ PluginManager.prototype.startPlugin = function (category, name) {
 				// Handle non-compliant onStart(): push an error message and disable plugin
 				self.coreCommand.pushToastMessage('error',name + " Plugin","This plugin has failing start routine. Please install updated version, or contact plugin developper");
 				self.logger.error("Plugin " + name + " does not return adequate promise from onStart: please update!");
-				self.disableAndStopPlugin (category,name);
 				myPromise = libQ.resolve();  // passing a fake promise to avoid crashes in new promise management
 			}
 
@@ -299,7 +297,6 @@ PluginManager.prototype.stopPlugin = function (category, name) {
 				// Handle non-compliant onStop(): push an error message and disable plugin
 				self.coreCommand.pushToastMessage('error',name + " Plugin","This plugin has failing stop routine. Please install updated version, or contact plugin developper");
 				self.logger.error("Plugin " + name + " does not return adequate promise from onStop: please update!");
-				self.disablePlugin (category,name);
 				myPromise = libQ.resolve();  // passing a fake promise to avoid crashes in new promise management
 			}
 			
--- a/app/pluginmanager.js
+++ b/app/pluginmanager.js
@@ -17,6 +17,8 @@ var variant = '';
 var device = '';
 
 module.exports = PluginManager;
+
+
 function PluginManager(ccommand, server) {
 	var self = this;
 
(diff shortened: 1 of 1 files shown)


fanciulli:
Hi,
Looks like a good work. What I am worried about at the moment (looking at the pr on the iPad) is when you replace the promise field in a defer, as in there following: 

defer.promise=plugin.onStart();

I would not follow this approach since the object defer has been created by the library Kew and you cannot know if there are some dependencies in it that you're breaking. I know that the code works as you report but we may find some issues in the future. 

Moreover in the pr description you said that if no promise is returned by the onVolumioStart a crash is expected. I would approach there point by avoiding any synchronization for that plugin. We have to start thinking that a lot of plugins will be done and some may not follow the APIs.

macmpi:
Hi, thanks for your feedback, I appreciate it.

> defer.promise=plugin.onStart();

Yes, that's a thing I was a bit unsure about.
Would you have another suggestions to pass plugin.onStart() promise back to caller level?
It is the same for launch & stop so the same method would be applied the same way.

> Moreover in the pr description you said that if no promise is returned by the onVolumioStart a crash is expected.

It’s currently the same behavior with any of onStart, onStop today.
This exception is something any developer will actually catch 100% of the time when developing the plugin: there is a javascript error message at runtime.
So, it’s never going to happen when user uses the plugin as developer always run its plugin at least once I guess...
Also catching the “bad practice” right-away is actually a good thing, as it avoids a new plugin interferes with all other installed plugins sequence management and creates unexpected bugs.
Don’t you agree?

macmpi:
@fanciulli : I fixed the issue(s) about
> defer.promise=plugin.onStart();

and tested accordingly. Please let me know if you think it is more suitable now, or not...

macmpi:
@fanciulli @volumio 

> Moreover in the pr description you said that if no promise is returned by the onVolumioStart a crash is expected.

Working on this part, to safeguard overall system operations against 3rdParty plugins not returning a promise from `onVolumioStart()`.
While preventing Volumio crashes and endless restarts, I was thinking to:
- disable such detected "faulty" plugin, and
- display a toastMessage to advise looking for an update or contact developer.

This way, it will avoid users' device to enter into crashing deadlock when updating Volumio with some older plugins. (BTW, many plugins are already fixed, so updating quickly will not be an issue).
Meanwhile, this will also promote conformance to developers for critical things such as plugins life-cycle (load/start/stop): currently this just goes unnoticed.
Not managing this life-cycle properly may lead to quite bogus situations (plugin may "start" before "load" is finished, very little respect of boot-order priority, etc...).

I have this proposed check working for `load` conformance (`onVolumioStart()`), but hold on submitting pending your views on this: is the 2-steps proposed approach OK with you?

Shall we do the same kind of "return promise protection" for `start` and `stop`? (this seems well respected currently for those two, as developers doc mentions it...)

volumio:
For this to be merged, we should make sure volumio properly starts even with malformed plugins: that's because if we release an update people would have non working systems if they did not update plugins

macmpi:
This is what I proposed in previous comment, but I'd like your views on the proposed automated "fail-safe" strategy before submitting that change: are you ok with "disabling such plugin" and "displaying a toastMessage to advise looking for an update or contact developer"?
There will be no crash, current users will be warned (all fine plugins update are available), and future developers will see this warning as one of the very first sanity check on their prototypes.
Do you agree?
I can submit changes early next week.

volumio:
I would avoid not starting the plugin. A warning on logs should be sufficient, what do you think?

macmpi:
Users would not notice log entry in most cases, and such "non-compliant" plugins may cause issues in the future (once load/start order is better sequenced).
It would be preferable user gets some clear message that updating affected plugin is recommended: toast warning seems relevant in that respect: it is particularly useful on manual ON triggering (toast not shown at automated boot start, as UI is not ready).
So we could do both: log & toast?

Disabling plugin is stricter I agree: though it might first seem "too-much", it won't be a major inconvenience as compliant plugins are/will be available on day-one (last remaining ones have pending PRs in volumio-plugins, other devs have fixed theirs,  only GPIO_Buttons still in PR).
With "disabled" enforcement, developers won't be able to ignore this requirement in their future developments: so mid-term, users won't suffer from it, it will be fixed during development.
Overall this ensure quickest & smooth transition.

macmpi:
No more crashes in case of non-promise-aware plugins.
Thanks for your comments: probably easier to discuss with code.

volumio:
@fanciulli what do you think?

fanciulli:
@volumio @macmpi  looks good to me.

volumio:
It can be merged. Before doing so I would however edit in a way that it displays the warning message, as it does now, but it starts the plugin (so do not disable).

macmpi:
ok, will make the change

macmpi:
@volumio done as requested.
Let me know if ok now.

volumio:
Very good, merging

macmpi:
Great, thanks.
Now one can look at:
- fix some out-of-promise threads in `i2s_dacs` and `system plugins` 
- make more deterministic `/volumio/app/index.js` startup sequence (ensuring plugins `starts` happen after all `loads`, & other operations sequence until startup tone)
- ensure load order by boot priority batches

PR #1170, opened 2017-05-08
Various fixes and enhancements for mpd/index.js

liberatorcnx:
Added 3 variables at start of file;

var tracknumbers = true/false - If set to 'true' then track numbers are
added to song titles.

var compilation = ['Various', 'VA', 'Various Artists'etc...] -
Identifies different strings which are used as AlbumArtist for
compilation Albums.

var artistsort = true/false - List artists by albumartist if set to true
or artist if set to false.

Fixed search to find albums which have duplicated names (Greatest Hits
etc...).

Added count of matches in search results.

Fixed adding album to Queue or Playlist which has name that is
duplicated (previously it added tracks from all albums with duplicate
name)

Fixed browsing by Genre to allow 'drilling down' correctly.

Probably a few other fixes I've forgotten......

### DIFF
--- a/app/plugins/music_service/mpd/index.js
+++ b/app/plugins/music_service/mpd/index.js
@@ -3031,6 +3031,7 @@ ControllerMpd.prototype.listGenre = function (curUri) {
     var defer = libQ.defer();
     var splitted=curUri.split('/');
     var genreName=nodetools.urlDecode(splitted[2]);
+    var genreArtist=nodetools.urlDecode(splitted[3]);
     var response={
         "navigation": {
             "lists": [
@@ -3070,7 +3071,13 @@ ControllerMpd.prototype.listGenre = function (curUri) {
     self.mpdReady
         .then(function() {
             var cmd = libMpd.cmd;
-            self.clientMpd.sendCommand(cmd("find genre \"" + genreName + "\"", []), function (err, msg) {
+            if (genreArtist != 'undefined') {
+            var findString = "find genre \"" + genreName + "\" artist \"" + genreArtist + "\" ";
+            }
+            else {
+                var findString = "find genre \"" + genreName + "\"";
+                }
+            self.clientMpd.sendCommand(cmd(findString, []), function (err, msg) {
                 var albums=[];
                 var albumsArt=[];
                 var artists=[];
--- a/app/plugins/music_service/mpd/index.js
+++ b/app/plugins/music_service/mpd/index.js
@@ -2413,10 +2413,17 @@ ControllerMpd.prototype.getMyCollectionStats = function () {
 
             self.clientMpd.sendCommand(cmd("list", ["album", "group", "albumartist"]), function (err, msg) {
                 if (!err) {
-                    var splittedAlbum = msg.split('\n').length;
+                    var splittedAlbum = msg.split('\n');
+					var albumsCount = 0;
+					for (var i = 0; i < splittedAlbum.length; i++) {
+					var line = splittedAlbum[i];
+					if (line.startsWith('Album:')) {
+						albumsCount++;
+						}
+					}
                     var response = {
                         artists: artistsCount,
-                        albums: (splittedAlbum - 1) / 2,
+                        albums: albumsCount,
                         songs: songsCount,
                         playtime: convertedSecs.hours + ':' + ('0' + convertedSecs.minutes).slice(-2) + ':' + ('0' + convertedSecs.seconds).slice(-2)
                     };
@@ -2428,8 +2435,8 @@ ControllerMpd.prototype.getMyCollectionStats = function () {
 
         }
 
-
     });
+    
     return defer.promise;
 
 };
--- a/app/plugins/music_service/mpd/index.js
+++ b/app/plugins/music_service/mpd/index.js
@@ -1590,58 +1590,6 @@ ControllerMpd.prototype.setAdditionalConf = function (type, controller, data) {
 	return self.commandRouter.executeOnPlugin(type, controller, 'setConfigParam', data);
 };
 
-ControllerMpd.prototype.getMyCollectionStats = function () {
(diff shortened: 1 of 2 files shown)


liberatorcnx:
The additional variables added will require some additions to the UI to allow the user to set the appropriate values.  I don't have any knowledge of the UI but I'm sure it will be simple for someone with this.... :-)

volumio:
Sorry if I did not come with a feedback. I need lot of time to test it, and also I need to add the UI part of it. Give me some days and I'll get back to you...

app/plugins/music_service/mpd/index.js
```diff
@@ -13,6 +13,12 @@ var mm = require('musicmetadata');
 var os = require('os');
 var execSync = require('child_process').execSync;
 var ignoreupdate = false;
+//tracknumbers variable below adds track numbers to titles if set to true. Set to false for normal behavour.
+var tracknumbers = true;
+//compilation array below adds different strings used to describe albumartist in compilations or 'multiple artist' albums
+var compilation = ['Various', 'Various Artists', 'VA', 'various', 'va', 'various artists', 'Bryan Ferry & Roxy Music', 'Fleetwood Mac and Christine Perfect'];
```
volumio: Please remove 'Bryan Ferry & Roxy Music', 'Fleetwood Mac and Christine Perfect'

app/plugins/music_service/mpd/index.js
```diff
@@ -1353,9 +1382,13 @@ ControllerMpd.prototype.search = function (query) {
...
 		{
-		    list=[
+			var artistdesc = "artist"
+			if (artistcount > 1) artistdesc = "artists";
+			list=[
                 {
-                    "title": self.commandRouter.getI18nString('COMMON.SEARCH_ARTIST_SECTION'),
+//                "title": self.commandRouter.getI18nString('COMMON.SEARCH_ARTIST_SECTION'),
+//title description below will need getI18nString changed....
+					"title": "Found " + artistcount + " " + artistdesc + " matching '" + query.value +"'",
```
volumio: Could you please change to translated strings?

app/plugins/music_service/mpd/index.js
liberatorcnx: Ohh - didn't realise I had left those in :-(

liberatorcnx:
Added the required updates.

volumio:
It has been merged... don't know why its still open
BTW, good work! Few errors here and there (shows that you did not test it extesively, please do it next time) but good one

volumio:
We have regressions, see https://volumio.org/forum/185-browse-album-play-t6793.html

liberatorcnx:
@volumio - not sure how it happened but my copy of this file already has the urldecoded line...  It only affects genres with spaces in the name - hence the fact that some worked OK and others didn't and it wasn't obvious when testing .  Please check to confirm.  The original PR did nothing to change artist artwork - maybe related to the recent album-art merge or maybe that genre navigation uses artist and not albumartist....

liberatorcnx:
@volumio - I think the PR was staying listed because my previous change (Added required 'Track' (singular)) was not commited - I think you maybe added it manually.  It needs to be cleared but not sure how?

volumio:
That's funny, but this reports my number of albums at: 267.5 

Probably the best way to test this would be setting a folder or USB drives with a known number of artists, songs and albums and see if they do get the right numbers

liberatorcnx:
Strange....  mpd returns 2 lines per album (plus a newline at the end) so the result of (splittedAlbum - 1) / 2 should be a whole number!

liberatorcnx:
Ohh - I think you have some bad albumartist tags...  I was able to reproduce this by adding two albumartists to a track - 'albumartist = The Rolling Stones//The Rolling Stones'.  When I look at the track in mp3tag it shows the albumartist as just 'The Rolling Stones' but when I look at the output from mpd (I wrote the output to file) it looks like this;
```
Album: Ruby Tuesday (Maxi CD Single)
AlbumArtist: The Rolling Stones
AlbumArtist: The Rolling Stones
```
So now I have 3 lines for that album...

Also, mp3tag shows the track albumartist correctly as 'The Rolling Stones//The Rolling Stones' in the side pane...

liberatorcnx:
If we want to allow for bad albumartist tags then I think we should just count the lines starting with 'Album:'  This seems to work:

```
            self.clientMpd.sendCommand(cmd("list", ["album", "group", "albumartist"]), function (err, msg) {
                if (!err) {
                    var splittedAlbum = msg.split('\n');
					var albumsCount = 0;
					for (var i = 0; i < splittedAlbum.length; i++) {
					var line = splittedAlbum[i];
					if (line.startsWith('Album:')) {
						albumsCount++;
						}
					}
					var response = {
                        artists: artistsCount,
                        albums: albumsCount,
                        songs: songsCount,
                        playtime: convertedSecs.hours + ':' + ('0' + convertedSecs.minutes).slice(-2) + ':' + ('0' + convertedSecs.seconds).slice(-2)
                    };
                }

                defer.resolve(response);

            });
```
Please test it!

volumio:
Ok the fix proposed here for album count works as expected. 

What still needs to be fixed:
- Play doesn't work anymore for many albums while browsing within a Genre.
- Add to playlist an artist from genre adds the whole genre

liberatorcnx:
@volumio - please give me some data on albums not playing.  Is this after you select a genre then try to play an album?  I am unable to reproduce this.  It should (and does for me) play all the tracks with a matching genre and artist in that album.  It does not play the whole album by design.

Add artist to playlist doesn't work and can never have worked before.  playlistManager calls mpd/index.js listGenre which has no code for drilling down to artist - it can only list all tracks matching a genre.  Regardless, I will look at this and add the required code :-)

volumio:
Good one, tested

Issue #1192, opened 2017-05-28
[RPi] Plugin Installation Breaks when the OS Password is Changed


mkonersmann:
Version: 2.175
Platform: Raspberry Pi

I installed the Raspberry Pi image and changed the user password for the OS via SSH.
Now the installation of the Youtube Plugin fails.

The "details" section in the UI did not help:
```
Downloading plugin at http://volumio.github.io/volumio-plugins/plugins/volumio/armhf/music_service/youtube/youtube.zip
END DOWNLOAD: http://volumio.github.io/volumio-plugins/plugins/volumio/armhf/music_service/youtube/youtube.zip
Creating folder on disk
Unpacking plugin
Checking for duplicate plugin
Copying Plugin into location
Installing dependencies
The following error occurred when installing the plugin: Error
```

The log says:
```
volumio@volumio:~$ tail -f /var/log/volumio.log
2017-05-28T08:41:00.371Z - info: Downloading plugin at http://volumio.github.io/volumio-plugins/plugins/volumio/armhf/music_service/youtube/youtube.zip
2017-05-28T08:41:01.793Z - info: END DOWNLOAD: http://volumio.github.io/volumio-plugins/plugins/volumio/armhf/music_service/youtube/youtube.zip
2017-05-28T08:41:14.350Z - info: Checking if plugin already exists
2017-05-28T08:41:14.355Z - info: Rename folder
2017-05-28T08:41:17.669Z - info: Move to category
2017-05-28T08:41:46.801Z - info: Checking if install.sh is present
2017-05-28T08:41:46.806Z - info: Executing install.sh
2017-05-28T08:41:48.692Z - info: Install script return the error Error: Command failed: echo volumio | sudo -S sh /data/plugins//music_service/youtube/install.sh > /tmp/installog
[sudo] password for volumio: Sorry, try again.
[sudo] password for volumio: 
sudo: 1 incorrect password attempt

2017-05-28T08:41:48.696Z - info: An error occurred installing the plugin. Rolling back config
2017-05-28T08:41:48.696Z - info: Plugin folders cleanup
2017-05-28T08:41:48.697Z - info: Scanning into folder /volumio/app/plugins/
2017-05-28T08:41:48.697Z - info: Scanning category audio_interface
2017-05-28T08:41:48.698Z - info: Scanning category miscellanea
2017-05-28T08:41:48.699Z - info: Scanning category music_service
2017-05-28T08:41:48.700Z - info: Scanning category plugins.json
[...]

Issue #1166, opened 2017-05-07
Https Login


NicoHood:
It would be nice to have the web frontend available with https. This way all data is encrypted, especially the login, for example if you configure spotify. A self signed https cert would be okay. You can trust it and then the traffic is encrypted.

volumio:
That makes no sense, sorry. There is no way we can have https on local volumio, for various reasons, the first of which is, if self singned you will get a ton of warnings by every browser

koenbuyens:
It would be great if 
- all the references to http:// in the code-base are replaced with protocol-relative URs (i.e. URIs starting with // instead of http:// ). 
- replacing explicit references to the IP address with the hostname of the URI
This would enable me to use nginx as reverse-proxy.

![image](https://user-images.githubusercontent.com/17049884/57990199-53a9ad00-7a72-11e9-97ec-11a457b9182b.png)

volumio:
Https over local network is not possible.

koenbuyens:
I am running https on my local network.

Koen

PR #1193, opened 2017-05-30
#1192 Enable plugin installation with changed OS password

mkonersmann:
Platform: Raspberry Pi

A fix to #1192. Where the plugins could not be installed when the OS password changed.

Removes 'sudo' from the plugin installation script. 
Root seems not to be required by the installation scripts of current plugins.
This has the benefits that neither the program itself needs to know any standard or generated sudo password (which would leave the sudo password readable), nor has the UI request for the sudo password and send it through the (unencrypted) network.

### DIFF
--- a/app/pluginmanager.js
+++ b/app/pluginmanager.js
@@ -841,7 +841,7 @@ PluginManager.prototype.executeInstallationScript = function (folder) {
 		}
 		else {
 			self.logger.info("Executing install.sh");
-			exec('echo volumio | sudo -S sh ' + installScript+' > /tmp/installog', {uid:1000, gid:1000, maxBuffer: 2024000},function(error, stdout, stderr) {
+			exec('echo volumio | sh ' + installScript+' > /tmp/installog', {uid:1000, gid:1000, maxBuffer: 2024000},function(error, stdout, stderr) {
 				if (error!==undefined && error!==null) {
 					console.log(stdout);
 					console.log(stderr);


volumio:
Unfortunately we need root privileges to install plugins (mainly the touch display one). 
If we don't run this as sudo, we then need to set nopasswd for many operations (dpkg, mv, echo etc) that might create security issues. 
Do you have any suggestion? A possible solution might be grepping the user password from the passwd file

mkonersmann:
The password in `/etc/passwd` is hashed, so it won't help you, I guess. But I have a set of ideas:

In #1192 you said
> we would like to change default pw for user volumio

You could generate a random password on the first start and store it - readable by the user 'volumio' only - in a configuration file. But then how do you make this available to the installing user, without making it readable to everyone? Because that would meant everyone in my network has effectively root access to my Pi.

Or you could ask for the password in the GUI upon plugin installation (perhaps only if it is really required by the plugin), but then I would really encourage you to make HTTPS available (#1166).

BTW: I was actually able to install the touch plugin. I cannot test if it is working though, because I have no touch screen here.
Is it the only plugin that requires root for installation? You could install it as default in all platform builds, which can make use of it..

volumio:
- Cannot have https on local volumios
- Cannot add plugin for HDMI output (image will be too big for PI)

mkonersmann:
As a last resort, you could go on using the standard password, but provide a UI for setting a (non-standard) system password (e.g. in the system configuration page).
The password could be stored readable to the user 'volumio' only in a configuration file. 

Not as elegant as I would like it, but it should work. Especially considering that you thought about changing the password anyway.

volumio:
You're right... We'll go with that!

Issue #1182, opened 2017-05-17
Add to playlist doesn't add subfolders tracks


Arthurin:
Everything is in the title! Using last release of Volumio on Rpi, tracks are stored on an USB stick.

liberatorcnx:
Can you give more information?  I tried adding various folders to a playlist, including my 'Various Artists' album folder and it added every track from every sub-folder (30+ albums) to a playlist.

Here's what I did:
Created a playlist with all of the tracks from an album.

Browsed to Albums, then added the folder 'Various Artists' to the playlist I created above - works fine and added all tracks from all 'Various Artists' albums to the playlist.

Browsed by Artist add added all albums by Artist X to playlist - works fine and added all of Artist X albums and tracks to the playlist.

If you can describe exactly what you did maybe someone can help.

Arthurin:
Sure let's detail the steps :
- Browse > Music Library > USB > select specific folder (containing multiple albums) options > Add to playlist > playlist named "Soundtracks"
![17-05-18 at 01 38](https://cloud.githubusercontent.com/assets/6535135/26180588/ae86a76c-3b6a-11e7-89cb-22056a7f25de.JPG)
No error during the operation and I even got the toast message telling me it worked, but when I look at the playlist it's empty, or even better I checked the websocket communication
![17-05-17 at 17 29](https://cloud.githubusercontent.com/assets/6535135/26180541/6dcdb9fe-3b6a-11e7-8a8e-88017e93ea50.JPG)
As you can see the pushAddToPlaylist is empty ?!

liberatorcnx:
Hmm - it seems that this feature has never been written.  In the relevant section of playlistManager.js it says  'TODO: Deal with folders' .....  It is not dissimilar to adding to queue which does work so it may not be too difficult.  I'm not sure if anyone is already working on this?  @volumio ?

volumio:
Wanna help on this?

liberatorcnx:
@volumio - Are you talking to me?  I will have a look...  No promises (I can only work synchronously) !!!!

volumio:
Then you might want to dig into promises... It's a hell at first sight, but after it its just fantastic...

liberatorcnx:
Sorry - you missed my joke @volumio -  Irish humour (No promises means I don't guarantee (promise) I can fix it).  Anyway I will have a look and see if I can come up with a workable solution.

volumio:
[...]

PR #1209, opened 2017-06-09
Fix Issue #1182

liberatorcnx:
Add function listallFolder to mpd/index.js to list all songs from a
folder including sub-folders.
Include call to listallFolder function and additional code in
playlistManager.js


liberatorcnx:
@volumio - please check this - I had some problems with Github. I only want to merge the last one - Fix Issue #1182

PR #1211, opened 2017-06-10
improved promise management around sync/push state

nooogent:
The idea behind these changes is to always return a promise for pushState and syncState. Potential benefits include improved stability around state changes and more efficient resource management (I have no evidence of these!).

I've tested forward, back, pause, play, seek in normal playback mode and random mode. Adding favourites and displaying favourite status seems ok too.


volumio:
In fact this has some performance handicap since before resolving the promise of push state you always checkfavourites. I find this approach however interesting... thanks

Issue #904, opened 2016-12-04
Fix kernelsource 


volumio:
https://volumio.org/forum/volumio-and-kernelsource-t5286.html#p24876

xipmix:
Is it not better to fix the sudoers file?  '/bin/systemctl' is in the NOPASSWD: list already so the 'echo' and the sudo -S seem unnecessary. Also the target scripts all have the right ```#!``` line so all that should be needed is ```chmod a+x``` ... but the execute bit is already set.

I prepared a patch but don't want to bother you with a PR if you are on it already.

volumio:
Please send, it will be useful!

xipmix:
I got a bit overexcited and gave kernelsource.sh a working over. See [kernelsource-part2](https://github.com/xipmix/Volumio2/tree/kernelsource-part2). It builds on the changes in the PR already sent.

volumio:
Ehehe, Ok I will test asap! thx!

xipmix:
I found some problems; calling pull.sh and kernelsource.sh needs sudo -S or changes to the sudoers file. The first can be handled by not requiring root permissions. The second needs [this](https://github.com/xipmix/Build/tree/kernelsource-part2)

PR #1216, opened 2017-06-11
volumio_command_line_client: Use POSIX syntax for functions

taudac:
The traditional bourne shell (sh) does not support the 'function foo { ... }'
syntax, use the POSIX standard syntax, 'foo() { ... }', to improve portability.
The later syntax works with bash, sh and many other standard shells.

Fixes issue #904.

### DIFF
--- a/app/asd.sh
+++ b/app/asd.sh
@@ -34,23 +34,23 @@ case "$1" in
 esac
 
 
-function start {
+start() {
 systemctl start volumio.service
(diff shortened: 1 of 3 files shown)


volumio:
Thanks! 

PS: Did you see the volumionext branch? If TAUDAC is supported in 4.9.31 kernel, then we might add it to dacs.json

PR #1224, opened 2017-06-13
Czech translation update

stranskj:
Added translation of new library settings.

Removed duplicate entry TRACK in english translation


volumio:
Thanks! Since you're there, check the new stuff in index.js , on DSD section...

stranskj:
Not yet, tomorrow :-)

13. června 2017 22:56:35 SELČ, Volumio <notifications@github.com> napsal:

stranskj:
Could you please be more specific, what you mean?

volumio:
See here:
https://github.com/volumio/Volumio2/commit/4d444fc061c2827679db6d4dfc7d6142ff3fd097

there is a new DSD section which requires updated strings, in PLAYBACK_OPTIONS

stranskj:
Ah, I see. Thanks for noting. I will add it in the evening.
Btw. It would be interesting to have some communication channel to translators to notify them with such news...

volumio:
Yes, you're right. I need to come up with a solution for that...

stranskj:
Two ideas on noting translators:
1) Make a forum thread (or use the one with instructions), which translators could mark to "Follow" and let know such news there.
2) Make a "permanent" Issue on Github, and translators can follow this issue for news.

PR #1227, opened 2017-06-14
Czech translation update

stranskj:
DSD related settings


stranskj:
There is additional  typographic correction in English version (space before comma)

Issue #1220, opened 2017-06-12
Plugin manager: unziping during install over-pressures memory to death


macmpi:
It very much seems unzipping plugin archive with `decompress-zip` can easily generate critical low-memory conditions and also gets `node` process to take nearly all CPU.
This can be seen on lower specked Pis (PiZero for instance) and big plugins like `Youtube`.
Sometimes install will just crash after being blocked for a while right after `Creating folder on disk `message.

Maybe some [other zip deflating node package](https://www.npmjs.com/search?q=zip%20decompress&page=1&ranking=optimal) may have better memory & CPU management (or deflating could be done natively)?
BTW deflating directly into `/data/plugins/...` would also save RAM in `/tmp`, and more aggressive intermediate files cleanup strategies in `/tmp` might also help... (as pointed [here](https://github.com/volumio/Volumio2/issues/847#issuecomment-261084732))

sample log below:
```
Jun 11 13:11:14 volumio volumio[910]: info: Downloading plugin at http://volumio.github.io/volumio-plugins/plugins/volumio/armhf/music_service/youtube/youtube.zip
Jun 11 13:11:19 volumio volumio[910]: info: END DOWNLOAD: http://volumio.github.io/volumio-plugins/plugins/volumio/armhf/music_service/youtube/youtube.zip
Jun 11 13:12:38 volumio volumio[910]: info: [1497186758531] ControllerMpd::pushError
Jun 11 13:12:38 volumio volumio[910]: info:  Error: spawn ENOMEM
Jun 11 13:12:38 volumio volumio[910]: at exports._errnoException (util.js:1018:11)
Jun 11 13:12:38 volumio volumio[910]: at ChildProcess.spawn (internal/child_process.js:319:11)
Jun 11 13:12:38 volumio volumio[910]: at exports.spawn (child_process.js:378:9)
Jun 11 13:12:38 volumio volumio[910]: at Object.exports.execFile (child_process.js:143:15)
Jun 11 13:12:38 volumio volumio[910]: at Object.exports.exec (child_process.js:103:18)
Jun 11 13:12:38 volumio volumio[910]: at Object.status (/volumio/node_modules/wireless-tools/ifconfig.js:192:10)
Jun 11 13:12:38 volumio volumio[910]: at ControllerVolumioDiscovery.getDevices (/volumio/app/plugins/system_controller/volumiodiscovery/index.js:378:14)
[...]

PR #1231, opened 2017-06-16
Optimize plugin install/update

macmpi:
plugin install eventually does some quite intensive file I/O operations for deflating and moving directories around.
This PR replaces `decompress-zip` by native `miniunzip` and use linux `mv` operations.
Installs are now much faster, but more importantly much more RAM efficient.
`/tmp` is only used for first download, deflating happens in storage location, hence saving RAM (intermediate files are cleaned as soon as possible too).

Notes:
- `decompress-zip` can be removed from Volumio2 modules (not used elsewhere)
- `minizip` must be added to Build as per [relevant PR](https://github.com/volumio/Build/pull/207) (used `minizip` as `unzip` is overflowing `stdout`)

fixes https://github.com/volumio/Volumio2/issues/1220

### DIFF
--- a/app/pluginmanager.js
+++ b/app/pluginmanager.js
@@ -6,7 +6,6 @@ var libFast = require('fast.js');
 var S = require('string');
 var vconf=require('v-conf');
 var libQ=require('kew');
-var DecompressZip = require('decompress-zip');
 var http = require('http');
 var execSync = require('child_process').execSync;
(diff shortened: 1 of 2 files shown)


macmpi:
Should be complete now.
Thanks for feedback.

volumio:
Well done!

PR #1234, opened 2017-06-22
Update strings_zh_TW.json

PeterDaveHello:


volumio:
Thanks! If you notice, there are a lot of strings missing from english translation file:
https://github.com/volumio/Volumio2/blob/master/app/i18n/strings_en.json

If you could add those, that would be fantastic...

PeterDaveHello:
Will do later.

PR #1235, opened 2017-06-25
Shuffle Playlist improvment

abrarmahmood:
No longer uses Math.random, so the same track will not be played twice in the 'underlying' shuffled playlist.


volumio:
Random works actually better. But there are some issues:
- Clear queue, add four songs. When clicking the first one, track number 4 plays
- Enable repeat, when pressing next playback stops (it should go to next random track)

abrarmahmood:
I have resolved the first issue you have mentioned, however I cannot replicate the second issue?

volumio:
Yep, I've not been able to replicate the second bug, nevermind.

More feedback:
- Ensure that next song is not the song that's just finished (sometimes it happens)
- This one is trickier: if random is enabled and repeat to off or undefined, we need to do random on the queue until all songs in queue are played and then stop. If repeat is on, then we just shuffle as we're doing now. 

In any case, great work. It works like a charm!

abrarmahmood:
I haven now fixed both of the issues you have pointed out. Surprisingly, point 1 was more tricky than point two!

volumio:
Sorry, I've been super busy... Was able to give it a look just now. Two comments:
- Could you please take care of conflicts ? (sorry about that..)
- I notice you set a 500ms timeout, that's not the ideal thing to do in this case. Why you do it?

abrarmahmood:
I did not add the timeout, it was part of the existing code. I have now resolved the merge conflicts.

PR #1247, opened 2017-07-04
Reduce usage of sudo -S

xipmix:
Analysis of the code base finds:
 9 instances in app/plugins/system_controller/volumio_command_line_client/commands/kernelsource.sh
 6 in app/plugins/system_controller/volumio_command_line_client/volumio.sh
 2 more instances in app/pluginmanager.js.

This series expects volumio/Build/PR#222 has been applied to /etc/sudoers. It converts kernelsource.sh to only be runnable by root and volumio.sh to run both commands/kernelsource.sh and commands/pull.sh without needing ```sudo -S```. The ```systemctl``` calls in volumio.sh are already runnable with the NOPASSWD: option, so they are converted as well.

These changes have been tested on rpi2 with version 2.201 and seem to work ok.

This series leaves the install.sh and uninstall.sh scripts run by the plugin manager to be addressed at a later time.


volumio:
This is very promising, but it requires a wider approach: we need to find a viable solution also for install.sh and uninstall.sh before moving to some direction

xipmix:
I quite agree. I have been thinking about that. The core issue is that the plugin install/uninstall scripts have names fixed by the plugin API but have an essentially arbitrary path prefixed to them;  just allowing ```sudo ./install.sh``` seems ... easy to game.

My suggestion is along these lines
 * in the sudoers file make three Command_Alias definitions
   * SYSTEM system commands needed at runtime
   * BUILD system commands needed during install/build
   * PLUGIN plugin install/uninstall scripts

 * The first two can be automatically compiled with a program that finds all the sudo invocations in the code base.  I have a hack that seems to do this but it looks brittle. To make this tractable and reliable some conventions about sudo calls might be needed. Manual review would be required anyway.

 * The last group of commands can be changed during runtime. I thought this could be done by adding a script (tentatively named 'pluginsudo') that has a fixed path (installed as part of the system build). It would be the first item in the PLUGIN command alias. It gets called during plugin install/uninstall and its task is to add or remove the paths for the plugin's install/uninstall scripts to the PLUGIN alias. It should of course check the resulting file with ```visudo -c``` before installing it as the new ```/etc/sudoers```.

This still allows the install/uninstall scripts the capacity to run arbitrary code, just like they have now. But that should only be possible while they are installed (uninstall removes the path to the plugin's install/uninstall scripts). I don't think there is any checksumming or signing of plugin packages; that might be a worthy future goal.

ashthespy:
@xipmix Could you open up a new issue with your proposed solution for the plugin installer? Would be easier to track/implement.  AFAIU, this is only blocking issue to let people change the default Volumio password?

xipmix:
I'm not sure I've followed your train of thought, please comment/correct me if I misunderstood.

 I am not sure if this is the only blocker, I would have to go back and reread the various issues that have been opened in the last couple of years.

ashthespy:
Hmm, from a quick look at the current master. 
```
app/pluginmanager.js
1310:  exec('echo volumio | sudo -S sh ' + installScript+' > /tmp/installog', {uid:1000, gid:1000, maxBuffer: 2024000},function(error, stdout, stderr) {
1339:  exec('echo volumio | sudo -S sh ' + installScript+' > /tmp/installog', {uid:1000, gid:1000, maxBuffer: 2024000},function(error, stdout, stderr) {

app/plugins/music_service/mpd/index.js
796:   execSync('echo "volumio" | sudo -S /bin/cp -f /usr/bin/mpdsacd /usr/bin/mpd', {
827:   execSync('echo "volumio" | sudo -S /bin/cp -f /usr/bin/mpdorig /usr/bin/mpd', {
```
Unless I am mistaken - these are the only remaining blockers for the hard-coded password right? - or are they for example other issues I am not seeing. I will read up and check back in the meantime.

Issue #1161, opened 2017-05-02
no picture in album view


Warter21:
Hi!

I have a strange problem. I browse the albums, sometime I cannot see the picture (folder.jpg) on the album view. But if I go into the albums any of these, the picture is magically appear. So I guess it is not a permission issue. Any idea how can I solve it?

Thanks,
Warter

Warter21:
maybe this causes the trouble:

ERROR getting albumart: JSON Error: Album not found for Infopath '/data/albumart/web/Elvis%20Presley/Viva%20Las%20Vegas%20-%20(Disk%202)/info.json'"

root@volumio:/data/albumart/web/Elvis%20Presley# ls -la
-rw-r--r--   1 volumio volumio 105297 Apr 23 21:41 dda6f14b-5c56-4407-8941-744747e4939b.png
-rw-r--r--   1 volumio volumio     63 Apr 23 21:41 info.json
drwxr-xr-x   2 volumio volumio   4096 Apr 23 20:46 Viva%20Las%20Vegas%20-%20(Disk%201)
drwxr-xr-x   2 volumio volumio   4096 Apr 23 20:46 Viva%20Las%20Vegas%20-%20(Disk%202)

root@volumio:/data/albumart/web/Elvis%20Presley# cat info.json
{
  "extralarge": "dda6f14b-5c56-4407-8941-744747e4939b.png"
}

xipmix:
Can't reproduce on 2.129. Can you write a bit more precisely the sequence you follow, eg
```Browse->Albums->[albumname]``` or ```Browse->Artists->[artist]->[albumname]```

Very handy if you can ```tail -f /var/log/volumio.log``` in a terminal window while you are doing this, so you see exactly when the error comes.

Warter21:
Yes, you are right. Moreover If I go to "Artists" and "Aerosmith", I can see the right pictures under "Albums" and "Tracks" too.

Logs:
--- Browse ---
[1494431867625] ---------------------------- Client requests Menu Items
2017-05-10T15:57:47.640Z - info: [1494431867635] Listing playlists
2017-05-10T15:57:47.650Z - info: CoreCommandRouter::executeOnPlugin: appearance , getUiSettings
2017-05-10T15:57:47.978Z - info: [1494431867973] ------------------------------ 348ms
2017-05-10T15:57:57.430Z - info: CoreCommandRouter::volumioGetBrowseSources
[...]

PR #1249, opened 2017-07-04
Fix issue #1161

liberatorcnx:
This fixes issue #1161

It will use any correctly names images inside the library album folder as album art and display them on the album list page.

On boot it will run the listAlbums function and store the result in cache.  On FIRST request to view albums it will read the cache album list file and copy any matching art files in the album folder to the volumio albumart folder and only download art for albums that have no album art file.

The cached result will be updated after a library update.

Images in the browser cache should be cleared before first running with this update.

Subsequent requests to view albums will be loaded from cache by volumio and the images will usually be loaded from the browser cache.


andreas-henning:
Hey liberatorcnx

I just wanted to say that I checked this out and it works perfectly! Thank you so much!!!

I own MANY albums where 'Web Album Art' lookup finds no album art at all, or worse, retrieves the wrong album art. Given the result (in my case no media organizer does automatic album art retrieval well) I would prefer no album art at all. Considering all the time I've invested to provide the correct album art for every album I own, that is just very painful to live with.

More importantly, my wife browses albums visually (not by album or artist name). For her this is a must-have feature. That 'Browse > Albums' can't display locally stored album art would have killed Volumio for us. That would have been a real shame, considering Volumio does so many other things so incredibly well :-/

I was about to move on, when I luckily found your contribution! So thank you again!

In my view, this should not only be possible in 'Browse > Albums' but also in 'Browse > Music Library > [Paths]'. If a directory contains cover.jpg or folder.jpg, it should be displayed there too.

Anyway, this certainly proves there are other ways to smoothly and quickly provide this feature without Volumio having to go as far as to recreate the entire music database itself. I hope we can rely on the team here to use this pragmatic solution. Even if it's not perfect from a technical/software-engineering point of view, it definitely solves a real problem in a way that works. Volumio is a far better product with this patch than without it.

volumio:
My opinion on this is: its not done properly in a technical way, but it works.
So I won't merge this one (it can create performances issues and high RAM usage). But I will take it as inspiration to solve the issue.
liberatorcnx showed its possbile, so let's make it

liberatorcnx:
I have closed the PR.  From the comments of @volumio - it looks like you have not really tested it.   There are no performance issues, only increased speed and fluidity.  With my (large) library the memory impact is <2k.   From previous issues and comments it seems that artwork is not considered important to this project which is why I spent a great deal of time coming up with a solution that works given the less than ideal tools available (no database).  As for the comment that 'it is not done properly in a technical way' - words fail me.

I will watch progress on this with interest.

app/plugins/user_interface/websocket/index.js
```diff
@@ -537,6 +537,12 @@ function InterfaceWebUI(context) {
 			});
 
 			connWebSocket.on('getBrowseSources', function (date) {
+				
+				//Refresh cacheAlbumList			
+				var mpdPlugin = self.commandRouter.pluginManager.getPlugin('music_service', 'mpd');
+				mpdPlugin.listAlbums();
```
volumio: This will be a problem, since this function is called every 4 seconds when user sits on my music page

volumio:
If you're looking to improve the project, you need to accept criticism from the mantainer, which has the full picture. 
In this project there were many times when we had technical discussions, and in the end the best implementations were originated by animated discussions. 
This is actually a good start, but its not ready to be merged, and needs  improvements and some rework. 
Mainly:
- If possible, we should avoid to ask users to clean their image cache
- Instead of storing large values in the memcache (node has already a big memory footprint, we should avoid making it bigger). A possible solution could be to start a process in the background after the db has been updated which links the proper images in the albumart folder. The best place to do it is:
https://github.com/volumio/Volumio2/blob/master/app/platformSpecific.js#L86
since from there we can tell when the db has been updated.
- The above could be extended to all albumarts, avoiding the long wait at first fetch.
- We should avoid to touch any browse related function, since we deal with long lists and we might have performances issues

This rework might be potentially one of the biggest improvements ever for Volumio, let's take the time and patience to do it proper.

andreas-henning:
@volumio, @liberatorcnx

I don't know if feedback on this is still appreciated, but I'll provide it just in case.

My music is distributed across three directories, stored on a USB drive, plugged directly into the RPi3.

As mentioned elsewhere, this patch changes the ordering of albums in 'Browse > Albums'. Pre-patch they are ordered by album-name. Post-patch the intent is for albums to be ordered by album-artist (I'm inferring that without really knowing). In my case they are actually sorted by directory and within that by album-artist.

That is likely unintentional.

volumio:
I spent some time in the weekend findin alternative ways of achieving the same result. In the end your implementation is better in any possible way. I made few fixes and merged it.

You were right and I were wrong, good work

liberatorcnx:
@volumio - You appear to have commented out the 'listAlbum()' command in websocket/index.js -
 connWebSocket.on('getBrowseSources)'.  You have left the following line 'self.logger.info('cacheAlbumList being updated');'  which of course will no longer be true!  You previously commented that 'this function is called every 4 seconds when user sits on my music page'.  I can find no evidence of that but maybe I am missing something?  Without this (or a replacement elsewhere) the cache will not be populated during boot.  Maybe you are happy with the additional delay when first running 'Browse Albums' but I would prefer it is this was completed during boot and the response to 'Browse Albums' was instant.

There is also still the issue of sorting the results.  The mpd command ' search album "" '  returns results in path order which equates to albumartist order with a single music source but this falls over if there are multiple sources (see comment from @andreas-henning above). 

The mpd documentation alludes to a 'sort' option (see [here](https://www.musicpd.org/doc/protocol/database.html)) but it doesn't work  We really need to find a way to sort the output from mpd to resolve this - I am looking at this now but would appreciate any input.

I also think a similar approach should be applied to Browse Artists and have something working but the problem is with albums that have multiple artists.  Any thoughts on this?

volumio:
@liberatorcnx 
- Yes I forgot to remove the log statement, will do now
- My comment about it running every 4 secs was wrong, I mistaken it for the function that retrieves music stats. 
- I removed from the websocket part because it should not be there: this gets called multiple times (every time you access with a client), so its not just called the first time. I tried to put it on Volumio start on mpd, but this was issue too early. I agree with you that we should populate the list on start, but we need to find the right way to do it (maybe just after the startup sound?) 

As for sorting, I agree with you. And the more I look at dealing with mpd to browse and get info of files, the more I think we should dump everything into our own DB, where we can do everything that we want. Or, maybe, directly accessing mpd db (however this might be dangerous).

I think this can be a great chance to add another albumart feature that we've been requested since long, but that I never did due to performance issues: add an option to retrieve albumarts from files metadata. What do you think?

andreas-henning:
Since you guys are already throwing around concepts relating to album art...

Any chance you could also display cover.jpg/folder.jpg in 'Browse > Music Library > [Paths]'?

This would be a huge help for households where more than one person uses Volumio, or for anyone using some kind of folder based organization scheme. In my case it would allow each family member to easily visually browse only their own albums.

As far as I can tell the same path and image caching concepts should apply.

andreas-henning:
> I think this can be a great chance to add another albumart feature that we've been requested since long, but that I never did due to performance issues: add an option to retrieve albumarts from files metadata. What do you think?

Using album art in ID3 tags is always messy, because the album art information is logically flawed to begin with (album related information embedded in track's ID3 tags isn't normalized).

I've seen so many people baffled by the album art some media players presents them with. For example, one media player displayed the first image embedded in the ID3 tag as the album art. The user thought this was wrong, and checked with another ID3 tagger, which suggested the image embedded in the ID3 tag was something entirely different. The user just failed to notice that the ID3 tag in fact contained multiple images, and both programs were assuming different ones to be the front cover.

The fact that tracks in completely different folders can, according to their ID3 tag, belong to the same album can also throw people off. When multiple tracks in the same folder contain different album art things get even thornier. In both cases,  what is effectively used for album art can seem arbitrary. I've also seen people complain about how poorly a media organizer rendered album art, only to later realize that although all tracks embedded the same image, one track embedded a low resolution version. That happened to be what was used, but the difference was impossible to notice using only the small image rendered by the person's tagging application.

I don't know what the best solution is. My point is only that this should also be viewed as a potential support issue in addition to it being a performance issue.

xipmix:
> Any chance you could also display cover.jpg/folder.jpg in 'Browse > Music Library > [Paths]'?
>
> This would be a huge help for households where more than one person uses Volumio, or for anyone using some kind of folder based organization scheme. In my case it would allow each family member to easily visually browse only their own albums.

While I understand what you are getting at here, does the following work for you:
adding the same share multiple times, with a different folder at the end of the path.
Example - the share is nasbox:/music. You could add
  * alice - nasbox:/music/alice's_tunes
  * bob - nasbox:/music/bobsobrill
  * clare - nasbox:/music/clarevoyant

and so on... This separates the collections at least but not, of course, within the Browse>Albums or Browse>Artist views. If you are keen on this feature can you open a separate issue please?

andreas-henning:
@xipmix

I'm unsure if you understood what I'm attempting to say. The NAS isn't involved here. Our music collection with over 600 Albums is stored on a 128GB USB drive plugged directly into the RPi3. The tracks are organized into a directory structure like this:

[USB drive root]
|- Bob
   |- [album 1..n]
|- Alice
   |- [album 1..n]
|- Clare
   |- [album 1..n]

For this reason each member's music collection is already separated. That's not the issue. The issue is that we can currently browse the entire collection visually in 'Browse > Albums' or we can browse sub-collections by folder structure but then only textually in 'Browse > Music Library > [Paths]'. Most members in my household browse their collections visually (not by artist names which they often don't know), so the second option doesn't really help them.

If album art is being cached as discussed here, using that to populate 'Browse > Music Library > [Paths]' with folder.jpg or cover.jpg images (where available) seems reasonable.

Am I wrong, or did I misunderstand you?

xipmix:
Thanks for clarifying things, I think I misunderstood before. So you are proposing to display albumart when browsing the music library "drives" for want of a better word. 'Paths' is apt. You are most interested in the USB drives that appear under /media/USB, but it seems to me this could extend to the /mnt/INTERNAL area,  or NAS drives under /mnt/NAS/, as well. 

This seems tractable but there may be complications. The issue as I understand it is that the current paths to the artwork are constructed from mpd queries for Artist and Album, completely independent of the path to the file on disk. So when browsing the '[Paths]' volumio must examine the file at /mnt/USB/Bob/JoeSchmoe/Greatest_Hits/song.flac and work out (reliably) that the list of corresponding album art images resides at /data/albumart/web/JoeSchmoe/Greatest_Hits/info.json (there may be multiple sizes, cached in the same location). I'm really not sure how difficult this is.

Ideally once the association is made it would be cached somehow, but to deal with things like drives mounted read-only the cache area would have to be somewhere under volumio's control ... /data/albumart/music_library/USB/<drivename> ? Implemented as a symbolic link to the artwork image perhaps. One would have to deal with broken symlinks and clearly define the desired behaviour when browsing-by-path before an albumart image is available; I suppose that should trigger a web lookup which places the image under /data/albumart/web/ and then makes the symlink?

volumio:
I've implemented albumart also on file browser today ;) 
It will show just local albumarts, which IMHO is the most logical thing to do since we're browsing per file...

andreas-henning:
^ EXCELLENT! Exactly what I was hoping for! Thank you so much!

liberatorcnx:
@volumio - I think we should try to avoid having to 'press a button' to update the cache.  The cache is already refreshed the first time 'List Albums' is selected after boot and whenever the mpd database is updated.  Nobody is going to boot Volumio, then go to 'Settings' My Music' to update the cache each time they boot.  There are easily predictable situations where the cache will need to be updated - basically when the mpd database changes and this is already updating the cache.  The only other issue is on boot. If it is a problem to run listAlbums on boot then would it not be possible to save the state of the cache when it is updated and reload it on boot?

I also think we need to implement this for listArtists and enable the display of users artist images (astist.jpg etc), including in file browse, which still downloads artist images the same as listArtists.  What do you think?

volumio:
@liberatorcnx I think you've misunderstood what clearing cache does in my commit:
- Remove disk cache on /data/albumart
- Change every albumart url in order to invalidate browser cache 

It's not related on the list caching, and I agree with you it would be silly to ask people to push a button. 

To do those operations previously you needed to connect via ssh, rm -rf /data/albumart/* and clear your browser cache ;)

I think you're idea is great for list artist. What do you think about embedded albumarts?

liberatorcnx:
@volumio - you're right - I misunderstood - sorry.  Deleting the disk cache is a good thing from a 'maintenance' viewpoint also as it will purge any deleted albums etc. which otherwise just sit there for ever.

Looking at the artist artwork got me wondering about artwork file names.  My artist artwork is all stored as 'artist.jpg' in the first level of the artist folder.  Many of my folders have multiple images there (fanart, logos etc) so it is important the Volumio knows which file to use.  For albums we build a predetermined array of file names to check for.  We would need to do the same for artists and second guess the file names people use.  What do you think about having a section in 'Settings' for Artwork and allow people to enter the file names that they use - the same as we do for Compilation Albums?  They could set this for Artists and Albums separately. 

There is also the issue of albums with more than one artist - not Compilation type albums but albums where maybe one track has artist set to 'albumartist feat. guestartist' or similar.  The only simple solution to this I can think of is that they should have images named exactly as per the artist name, so we use, say, folder.jpg for all tracks except where artist !== albumartist where we use 'artistname'.jpg. What do people think?

With regard to embedded albumarts - as you know they are really embedded track-arts.  It is possible for each track to have different artwork, including multiple images per track and there are no de-facto rules for how these are embedded.  They bloat the size of each track file and I'm not convinced this is worth doing or how we would use them.  I also think it will lead to a lot of questions and issues from a support viewpoint as mentioned earlier by @andreas-henning . The only thing worth considering in my opinion is maybe extracting the first embedded artwork (from the first track with embedded artwork) if there is no separate album artwork.  For me it is more useful to list artists and albums with missing artwork and then correct the problem by adding it.

andreas-henning:
> There is also the issue of albums with more than one artist - not Compilation type albums but albums where maybe one track has artist set to 'albumartist feat. guestartist' or similar. The only simple solution to this I can think of is that they should have images named exactly as per the artist name, so we use, say, folder.jpg for all tracks except where artist !== albumartist where we use 'artistname'.jpg. What do people think?

I take it you're using a directory structure like this:

```
[album-artist-1]
  folder.jpg
  [album-1]
    folder.jpg || cover.jpg
  [album-2]
    folder.jpg || cover.jpg
[album-artist-2]
  [album-1]
  etc.
```

This structure is well suited to managing any and all album related information. The top level directories are album artists and nothing more (it excludes performers or contributing artists). This structure also avoids dependencies between metadata and the filesystem, i.e. we can add/remove/modify any metadata without consequence, i.e. imagery will still display as expected solely based on the image file's location in the hierarchy. IMHO this is why it works.

Adding [artist-name].jpg image files to this structure breaks that. For artist imagery, it's likely impossible to avoid dependencies between metadata and names of objects on the filesystem. That's cumbersome enough. Unfortunately, adding artist imagery to this album-focused directory hierarchy will create situations where image files must be duplicated. This occurs whenever the same artist makes contributes to multiple tracks on different albums. I have a few dozen examples of that in my collection. It also occurs when a contributing artist (necessitating [artist-name].jpg) on one album is also an album artist for their own album (necessitating folder.jpg).

IMHO manually managing multiple copies of the same image files, sometimes with different names, and manually syncing metadata and filesystem dependencies is too cumbersome and error prone to make for a reliable solution in practice. IMHO this will only work for the most dedicated and persnickety music buffs with lots of free time, which is likely to be a very small minority.

Assuming Volumio doesn't intend to become a tagging application that automatically keeps advanced filesystem and metadata dependencies in sync, and assuming Volumio must provide a logically coherent and manageable solution for user supplied artist imagery, then I think the solution for that must be provided outside of that album-focused directory hierarchy.

I like your suggestion for handling embedded album imagery. It removes all the guesswork about where album art comes from, because anything that is extracted from metadata is written to the filesystem, making Volumio's selection transparent to the user. Now that Volumio's Music Library browser will also displays those images, using this approach, it would be simple for users to identify the source of any unexpected results and fix them.

I'd very much appreciate a similarly straight forward approach for user supplied artist imagery.

Would it maybe be a better idea to provide a single directory dedicated solely to collecting [artist].jpg files? If an [artist].jpg file is missing, Volumio might attempt to retrieve it from the web and write it to the directory. If that fails, Volumio might instead write an [artist].txt file to the directory as a placeholder.

Using a dedicated directory would avoid the need to duplicate image files. The placeholder file would make it easier for users to identify which artist imagery their collection is still missing and also make it easier to identify spelling mismatches. In a nutshell, this attempts to write files to the filesystem, with names based on existing metadata, so users can more easily identify what information Volumio has and what they must provide to match it, in a way that is similarly transparent to the user as the approach you suggested to handling embedded album imagery.

Just my two cents... I suspect this would probably be quite a lot of work to implement though :-/

volumio:
@liberatorcnx IMHO our first further step would be concentrating on embedded albumarts (which for some people are important) and I totally agree with your approach:
- If no albumart is found on the folder we take the first embedded albumart of the first item in the folder

This should be the optimal balance between performances and use of embedded albumart. Would you want to tackle this?

After this has been done, we can then take care of artist local albumart. But this might open a can of worm, as your situation might be quite "unique" and is definetely non-standard. As Andreas suggests, this might overcomplicate the artwork management...
But as usual, if you come up with a good implementation (like last ones), they will be integrated.
Let me know what you think

andreas-henning:
In 'Browse > Albums' the entries are still ordered by directory and within each directory by album artist. 

This was tested with the music library stored on a USB stick plugged directly into the RPi3.

Albums should be ordered simply by album artist.

liberatorcnx:
The ordering of all listing requests is dictated by mpd.  Although mpd alludes to a 'sort' option in reality it doesn't work.  The output from mpd is ordered by source then directory, so for people with a single source and conventional directory structure is fine, but not if you have multiple sources or use a different directory structure.  The only way I can think of to order the output of mpd would be to put it into an array and then sort the array before displaying.  I am not sure what the performance hit for that would be.  When I get some time I will look at that if nobody does so before.

PR #1262, opened 2017-07-12
pluginhelper: fix compress

macmpi:
Volumio does not include `zip` utility by default, so compress fails (unless added extra).
Since `minizip` is readily available, we use it instead.

### DIFF
--- a/pluginhelper.js
+++ b/pluginhelper.js
@@ -348,8 +348,8 @@ function zip(){
     try {
         if(fs.existsSync("node_modules")) {
             var package = fs.readJsonSync("package.json");
-            execSync("cd " + process.cwd() + " && /usr/bin/zip -r " +
-                package.name + ".zip *");
+            execSync("cd " + process.cwd() + " && /usr/bin/minizip -o -9 " +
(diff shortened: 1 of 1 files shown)


macmpi:
Done, tested fine.

volumio:
Very sensible, let me test and merge

pluginhelper.js
```diff
@@ -346,26 +346,20 @@ function refresh() {
...
             }
             catch (e){
                 console.log("Error installing node modules: " + e);
                 process.exit(1);
             }
-        }
+        }        
+        var package = fs.readJsonSync("package.json");
+        execSync("cd " + process.cwd() + " && /usr/bin/minizip -o -9 " +
+            package.name + ".zip $(find -type f -not -name " + package.name + ".zip -printf '%P ')");
```
volumio: Why this?

macmpi: `minizip` does not seem to support recursive directory exploration for compression: you need to feed it with explicit list of file paths. So used `find` to do it, taking care of excluding archive being created :)
Used this code for several releases of backup-restore plugin with no issue.

PR #1265, opened 2017-07-15
Fix special characters in artist name

Flibblebot:
Replaced nodetools.en/decodeURL with en/decodeURIComponent to encode URL allowed characters ( /?:@&=+$# ) - fixes issue with artist names such as AC/DC

Added codedArtistName to listAlbums function as this was missing

see [https://volumio.org/forum/album-view-not-converting-characters-t6989.html](url) for more info


volumio:
Thanks, good work!

PR #1285, opened 2017-08-01
Improve user input hygiene in logsubmit.js

xipmix:
@pilifsch found, to his/her confusion, that logsubmit does silly things when given a description containing parentheses. This series improves the quoting of the data from the /dev form so that should no longer happen.

Tested on stock 2.201 with just these changes.


volumio:
Thanks

Issue #1239, opened 2017-06-27
Software mixer not working: RpiB+ with Durio (hifiberry driver) on v2.201


heis2201:
I would love to use Software volume control (even though I loose some quality) but I can't get it working on Volumio 2.201.

Configuration:

> Output device: Hifiberry DAC
> I2S DAC: ON
> DAC model: Hifiberry DAC
> Mixer type: none/Software

When I set Mixer type to Software, and then Save, Volumio crashes and I have to power-off the raspi.

Anyone suggestions? Thanks!

volumio:
Which raspberry pi model do you have?

heis2201:
Sorry, forgot to mention: It's a B+

heis2201:
I have found the following information in my kernel log: What does it mean, and is it related?

```
[  +0.794184] snd_bcm2835: unknown parameter 'index' ignored
[  +0.013552] sd 0:0:0:0: Attached scsi generic sg0 type 0
[  +0.069926] snd-hifiberry-dac soc:sound: ASoC: CPU DAI (null) not registered
[  +0.008998] snd-hifiberry-dac soc:sound: snd_soc_register_card() failed: -517
[  +0.034636] sd 1:0:0:0: Attached scsi generic sg1 type 0
[  +0.148353] snd-hifiberry-dac soc:sound: ASoC: CPU DAI (null) not registered
[  +0.009035] snd-hifiberry-dac soc:sound: snd_soc_register_card() failed: -517
```

nilsbecker:
seems related to #1168 (although that is on rpi3 and started in 2.163)

riesch:
Exactly the same problem and behaviour with raspberry pi zero w and hifiberry miniamp. 
Output device: Hifiberry DAC
I2S DAC: ON
DAC model: Hifiberry DAC
Mixer type: none/Software

riesch:
Imo this is related to https://github.com/volumio/Volumio2/issues/1168

heis2201:
as @nilsbecker mentioned [above](https://github.com/volumio/Volumio2/issues/1239#issuecomment-311775552)...

heis2201:
anyone any ideas how to fix this?

heis2201:
I assume we can close this bug since https://github.com/volumio/Volumio2/pull/1300 has been merged into master (https://github.com/volumio/Volumio2/pull/1300#issuecomment-324387222)

Issue #1168, opened 2017-05-07
cannot set softvolume on 2.163 and 2.175 with hifiberry digi and rpi3 (regression)


nilsbecker:
when setting the mixer to software volume control, the web ui becomes unresponsive. ssh still works. after restart, the setting for volume conrol remains at 'undefined'

i have to use software volume control since i have a hifiberry digi+ digital output board without hardware mixer.

nilsbecker:
this is still the same. for information, my `/etc/asound.conf` exists but is empty and my `/etc/mpd.conf` has an entry `resampler` since i need resampling to 88.2. this is working in general -- except for the software mixer.

nilsbecker:
- bump -

i am still having this issue. i now re-flashed the image 2.175 to my rpi 3.

to reproduce:

flash the image.
boot the first time.
set the sound output to hifiberry digi.  (which does not have a hardware mixer). save.
wait 10min.
reboot. 
set the mixer to software, starting volume 50%. save setting. notification: settings saved.
navigate away from the playback options page.
navigate back to playback options: the playback options page does not display.
navigate to volumio.local: the page does not display. ssh still works.
reboot.
navigate to playback options page. the mixer is set to None.

Nicolas-94:
Hello 
+1 for this, I have a quite similar issue also on 2.141 (and next ones) with Raspberry 3, and 2 (B+), and an I-sabre DAC (setting:  hifiberry DAC). Once software mixer is selected, there is no sound anymore, and volumio displays on the webpage  an error 'can not open alsa output'.  Resetting software mixer to none does not  bring back the sound on. In order to have the sound, I have to  deactivate the DAC in settings, reboot, re-activate it and reboot again. 
I did not have any such issue before 2.141
I don't have a specific asound.conf or mpd.conf setting either. 
If you need more logs, please tell me which file to put here. 

(and high 5 to volumio team for the great job)

cortegedusage:
+1 for me for this one, also with the new release 2.185 my RPI3 with i2s AUDIOPHONICS I-Sabre V3 DAC ES9023 which worked with hififerry DAC, doesn't work when i select software mixer.

Before it did work.

willing to test extra if needed.

nilsbecker:
for me also it did work before 2.163

cortegedusage:
Still no resolution?

ArwynHarris:
[...]

PR #1288, opened 2017-08-02
fix softvolume bugs

luckynrslevin:
This fix is related to issue #1168 and #1239

1. Fix crash after changing one of the Volume Option parameters when Mixer Type is already set to „Software“
2. Fix bug setting Mixer Type from „Software“ to „None“

### DIFF
--- a/app/plugins/audio_interface/alsa_controller/index.js
+++ b/app/plugins/audio_interface/alsa_controller/index.js
@@ -649,13 +649,17 @@ ControllerAlsa.prototype.saveVolumeOptions = function (data) {
 	self.setConfigParam({key: 'mixer', value: data.mixer.value});
 	} else if (data.mixer_type.value === 'Software') {
 		var outdevice = self.config.get('outputdevice');
-		self.enableSoftMixer(outdevice);
+		if (outdevice != 'sofvolume'){
+			self.enableSoftMixer(outdevice);
+		}
 	} else if (data.mixer_type.value === 'None'){
 		self.setConfigParam({key: 'mixer', value: ''});
 		var outdevice = self.config.get('outputdevice');
 		if (outdevice === 'softvolume'){
             var outdevice = self.config.get('softvolumenumber');
             this.config.set('outputdevice', outdevice);
+            this.config.delete('softvolumenumber');
+            this.commandRouter.executeOnPlugin('music_service', 'mpd', 'restartMpd', '');
 		}
 		self.commandRouter.sharedVars.set('alsa.outputdevice', outdevice);
 		self.disableSoftMixer(outdevice);
--- a/app/plugins/audio_interface/alsa_controller/index.js
+++ b/app/plugins/audio_interface/alsa_controller/index.js
@@ -649,7 +649,7 @@ ControllerAlsa.prototype.saveVolumeOptions = function (data) {
 	self.setConfigParam({key: 'mixer', value: data.mixer.value});
 	} else if (data.mixer_type.value === 'Software') {
 		var outdevice = self.config.get('outputdevice');
-		if (outdevice != 'sofvolume'){
+		if (outdevice !== 'sofvolume'){
 			self.enableSoftMixer(outdevice);
 		}
 	} else if (data.mixer_type.value === 'None'){


luckynrslevin:
This fix is related to issue #1168 and #1239

My setup: RPI3 & HifiBerry DAC, Volumio VERSION: 2.246

Volume Options:
1. Fix crash after changing one of the Volume Option parameters when Mixer Type is already set to „Software“
Behavior:
Setting Mixer Type from „None“ to „Software“ works exactly once.
Changing one of the settings, like  „Default Startup Volume, Max Volume Level, …“ afterwards will crash the system completely after some seconds.
Reason:
It again will try to change the configuration to Software and therefore modify /etc/asound.conf, e.g. the card will be changed to softvolume instead of the number of the real audio device (e.g. in my case 1).
Fix:
I implemented the if statement to check if the device is already set to Software.

2. Fix bug setting Mixer Type from „Software“ to „None“
Behavior:
Setting the Mixer Type from Software to None will result in an alsa error message when trying to play music.
Reason:
Everything works fine except the „softvolumenumber“ parameter is not deleted from /data/configuration/audio_interface/alsa_controller/config.json and therefore in /etc/mpd.conf the device is not correctly reset to the hardware device (in my case: device hw:1,0).
Fix:
I added two lines to delete the parameter from the config and afterwards restart mpd to load the new device configuration.

xipmix:
At line 652
```
if (outdevice != 'sofvolume'){
```
do you mean
```
if (outdevice != 'softvolume'){
```

luckynrslevin:
I was thinking about it, but since I'm new to JavaScript & volumio I was not sure on the best practice. As I understood !== is verifying the type in addition to the value.
From a functional point of view both versions are working in my environment according to my tests.

xipmix:
I'm never sure about that either but given what the code around it
was doing I thought it worth asking the question. Cheers!

volumio:
This will not work because:
f (outdevice !== 'sofvolume')

is always true (typo: sofvolume should be softvolume)

Also, the deal here is getting softvol device up the first time, not after it has been created

luckynrslevin:
1) typo: sofvolume should be softvolume => yes, this is wrong.

2) if (outdevice !== 'sofvolume') is always true
Why, because outdevice is not of type String?
https://www.w3schools.com/code/tryit.asp?filename=FI9C5AVKIV97

3) Also, the deal here is getting softvol device up the first time, not after it has been created
I think I don't understand the comment, this is exactly the reason why I implemeted the if statement.  According to my remote debugging results (set breakpoint at line 618 and walk through the code) this code is also executed after the softvol device already has been created (e.g. if you only want to change the Default Startup Volume from 10 to 20) and therefore you end up with a not working /etc/ascound.conf.

Afterter switching Mixer Type to "Software" the first time my /etc/asound.conf looks like this and everything works fine:
>pcm.softvolume {
>    type             plug
>    slave.pcm       "softvol"
>}
>
>pcm.softvol {
>    type            softvol
>    slave {
>        **pcm         "plughw:1,0"**
>    }
>    control {
>        name        "SoftMaster"
>        **card        1**
>        device      0
>    }
>max_dB 0.0
>min_dB -50.0
>resolution 100
>}

If I now change the Default Startup Volume (e.g. from 10 to 20) an save again, afterwards /etc/asound.conf looks like this and you cannot play music any more:
>pcm.softvolume {
>    type             plug
>    slave.pcm       "softvol"
>}
>
>pcm.softvol {
>    type            softvol
>    slave {
>        **pcm         "plughw:softvolume,0"**
>    }
>    control {
>        name        "SoftMaster"
>        **card        softvolume**
>        device      0
>    }
>max_dB 0.0
>min_dB -50.0
>resolution 100
>} 

Changes of pcm and card marked in bold.

Any help / hint apreciated.
Thank you.

volumio:
I think you're actually right. Unfortunately I will be away from my studio this month and won't be able to test extensively. Is there someone willing to help test this?

xipmix:
> if (outdevice !== 'sofvolume') is always true
this was a suggestion without much analysis behind it, if != works then ok.

Let's try to clarify the situation: is outdevice always something other than a string? 
Perhaps it would be best to 'stringify' it before doing the comparison with 'softvolume' ?
That would make the code the clearest.

luckynrslevin:
1) != vs. !==
Reading some docs to my understanding !== could be a problem, if you use referenced types in volumio or the modules you use, e.g.:
>"abc" == new String("abc")    // true
>"abc" === new String("abc")   // false
see: https://stackoverflow.com/questions/359494/which-equals-operator-vs-should-be-used-in-javascript-comparisons#359509

Since we need to compare the value != should be ok from my point of vierw, but as already mentioned I do not have JavaScript programming experience.

2) Help testing the fix and regression tests

a) Testing if the fix solves the problem also for the other users
Since some people have reported the problem, I could ask them to test my fix. But how would they do this? I read the developer docs and searched the forum, but could not find a doc on how to test fixes in certain branches. The _volumio pull_ command to my understanding always pulls the current master from _github.com/volumio/Volumio2.git_. Would it be an apreciated enhancement to have a _volumino pull_ command, that has the option to pull a certain branch from another repository, e.g. _volumino pull github.com/some-user-forked-volumino/Volumio2.git -b branch-including-the-fix_ ?
In this case the users could flash their device and afterwards issue a special volumino pull command to pull the branch from my forked repository and give feedback on the bugfix.

b) Regression testing
I am willing to hlep, but I don't know how you usually do this.
Checking in my pull request, I saw some error reports from Travis CI. Is there something like a continuous integration and testing setup existing?

meaz:
hi @luckynrslevin 

I tried your index.js file by using ssh and changing the content of `/volumio/app/plugins/audio_interface/alsa_controller/index.js ` by the content of your file.

I just modifiy your file a line of your file, as suggested above:
`if (outdevice !== 'sofvolume')` by `if (outdevice !== 'softvolume')`

I rebooted volumio, but when I try to set the mixer to software volume control, the web ui still become unresponsive. Ssh still works. after restart, the setting for volume conrol remains at 'undefined'.

luckynrslevin:
Hi @meaz 
What is your setup? RPI?, DAC?

Can you provide results from the following commands:
`aplay -l`
`aplay -L`

can you post your current:
`/data/configuration/audio_interface/alsa_controller/config.json`
and
`/etc/asound.conf`
and the audio_output part  from your /etc/mpd.conf, something like
`…`
`audio_output {`
		`type		"alsa"`
		`name		"alsa"`
		`device		"????"`
		`dop			"no"`
`}`
`...`

luckynrslevin:
Please use the following procedure to test my fix:

1. Download the current volumio version (I used RPI VERSION: 2.246) and flash your SD Card
2. ssh to your volumio device
3. backup original file
`mv /volumio/app/plugins/audio_interface/alsa_controller/index.js /volumio/app/plugins/audio_interface/alsa_controller/index.js.ORG`
4. install my fix
`curl https://raw.githubusercontent.com/luckynrslevin/Volumio2/softvolume-fix/app/plugins/audio_interface/alsa_controller/index.js > /volumio/app/plugins/audio_interface/alsa_controller/index.js`
5. Make sure the file has correct owner:group and permissions
`ls -al /volumio/app/plugins/audio_interface/alsa_controller/index.js`
-rw-r--r-- 1 **volumio volumio 46440** Aug  5 19:57 /volumio/app/plugins/audio_interface/alsa_controller/index.js
6. reboot your device
7. In volumio Web Interface: _Playback Options => enable I2S DAC => Choose your DAC model_ (mine is HifiBerry DAC) and _save_
8. Reboot your device
9. Switch _Mixer_Type to "Software"_ and _save_

=> Afterwards playing sound including software volume control should hopefully work ...

meaz:
See my post here: #1168

broemels:
> Please use the following procedure to test my fix:
> 
> 1. Download the current volumio version (I used RPI VERSION: 2.246) and flash your SD Card
> 2. ssh to your volumio device
> 3. backup original file
>    `mv /volumio/app/plugins/audio_interface/alsa_controller/index.js /volumio/app/plugins/audio_interface/alsa_controller/index.js.ORG`
> 4. install my fix
>    `curl https://raw.githubusercontent.com/luckynrslevin/Volumio2/softvolume-fix/app/plugins/audio_interface/alsa_controller/index.js > /volumio/app/plugins/audio_interface/alsa_controller/index.js`
> 5. Make sure the file has correct owner:group and permissions
>    `ls -al /volumio/app/plugins/audio_interface/alsa_controller/index.js`
>    -rw-r--r-- 1 **volumio volumio 46440** Aug  5 19:57 /volumio/app/plugins/audio_interface/alsa_controller/index.js
> 6. reboot your device
> 7. In volumio Web Interface: _Playback Options => enable I2S DAC => Choose your DAC model_ (mine is HifiBerry DAC) and _save_
> 8. Reboot your device
> 9. Switch _Mixer_Type to "Software"_ and _save_
> 
> => Afterwards playing sound including software volume control should hopefully work ...

Hi [luckynrslevin](https://github.com/luckynrslevin),

I have a legacy Raspberry PI with the same issue but the link doesn't work anymore.
Is it possible to share what changed in /volumio/app/plugins/audio_interface/alsa_controller/index.js?

Thx

luckynrslevin:
@broemels I think you still can see the changes in the pull request:
https://github.com/volumio/Volumio2/pull/1300

BR

Issue #1237, opened 2017-06-26
Audio shuttering after update to 2.201 (Airplay)


pilifsch:
Hi guys,

I have a problem after last update. I updated Volume from something about v2.110 (can't remember exactly) to the newest version and I am experiencing audio shuttering every time while playing over airplay. This is not happening while playing webradio. I have such issue for the first time on Volumio 2.
I tried increasing buffer size to 12 MB and fiddling with other options (DSD playback, audio resampling) but with no luck. Can you please help me figure out what is wrong?

Thanks,
Philip

heis2201:
Same here.

pilifsch:
Anybody can help with resolving this issue?

jamesw4:
Same issue here aswell.

volumio:
Guys can you please share more infos about your setup?

pilifsch:
Sure I can.
As for hardware part. I have RPi 1 model B with WiFi USB dongle and external DAC connected through USB too. RPi connects to my home network using WiFi only. Power supply is efficient (2.4 A).
OS that I use is of course Volumio OS. However I have installed additional software like Homebridge and Mosquitto for my home automation.

I should mention that I have complete backup of state before update (created image of SD card). By that I can reproduce this issue. Let me know if we can use it somehow to find the cause of this problem.

empossible:
I've got the same issues (crackling, popping, disconnecting entirely). Mine is on a fresh install of 2.201 after formatting the SD card.  I did this twice with the same results.  I'm using a RPi 3 over WiFi, with a CanaKit power supply.  

After a third format, I tried the Volumio Hotspot.  That has resulted in to issues so far (about 40 minutes of playback).  When I switched back to my WiFi network, all the issues immediately return.

I would say it's an issue with my WiFi network, but I've also run RuneAudio on the same device and it had no trouble with Airplay over my WiFi network.

volumio:
Can you please add:
audio_backend_buffer_desired_length_in_seconds = 0.30;

in the alsa section of /volumio/app/plugins/music_service/airplay_emulation/shairport-sync.conf.tmpl

and reboot? does it help?

empossible:
Ok, I rebooted from the WebUI and the same issue arose immediately.  I tried again a few minutes later without rebooting, and I received static followed by a crash and reboot.  After the crash and reboot boot, it worked!  For like, 30 seconds.  Subsequent reboots resulted in the static, popping, and audio dropping immediately on playback.

[...]

PR #1292, opened 2017-08-05
More logsubmit work

xipmix:
The fix I submitted a few days ago did not handle an important case - single quotes in the description. I've fixed that now, in the same way as we did in networkfs. The fix has been tested, see eg pBwpQAT.

While I was there I noticed the catch() clauses were using what seemed to be the wrong name for the error object.

Finally as described in #1237 I found it may be useful to clean up the data sent to socket.on('pushSendBugReport` before trying to parse it into json.

Comments and further testing welcome. Tested on 2.201 plus these patches.

### DIFF
--- a/app/plugins/system_controller/system/index.js
+++ b/app/plugins/system_controller/system/index.js
@@ -391,8 +391,18 @@ ControllerSystem.prototype.sendBugReport = function (message) {
 		message.text = 'No info available';
 	}
 	fs.appendFileSync('/tmp/logfields', 'Description="' + message.text + '"\r\n');
-    // Must single-quote the message or the shell may interpret it and crash.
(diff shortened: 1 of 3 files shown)


volumio:
Well done!

volumio:
@xipmix we still have passwords exposed for NAS mounts... I think we should fix this asap... any idea?

xipmix:
can you point me to or send an example, I am happy to add further checks

volumio:
You can try adding a NAS drive with fake user and password credentials, and then submit logs. They are exposed on the mount command

xipmix:
Thanks, it's late here... I pushed an extra commit to this PR. Is that ok or should I submit separately?

volumio:
Yes do it separately, I'll look at logsubmit3

xipmix:
It's going to be a little while. There is something up with the behaviour on the command line vs within logsubmit.js. The current logsubmit3 branch I have removes everything after ```/bin/``` on the offending line.

PR #1294, opened 2017-08-06
Fix spelling of 'Cannot'

xipmix:
I know it's trivial but couldn't keep looking at it...


volumio:
It was trivial, but still a mistake ;) thanks

PR #1300, opened 2017-08-08
Softvolume bugfix - issues #1168 & #1239

luckynrslevin:
This fix is related to the issues #1168 and #1239 and to my former pull request #1288.
The fix is successfully tested by me and some other users reporting the problem (see #1168).
If I can support in additional regression testing, just give me a hint.


luckynrslevin:
Also related to issue #1304

volumio:
Tested extensively and it works great, thanks

PR #1302, opened 2017-08-10
pluginhelper: move package.zip out of source directory

macmpi:
`cp` command keeps a copy of plugin archive within source directory: at next submit this zip file will be part of commit in source directory, which is not desirable (plugin archive is submitted elsewhere).
We could do this differently (by having zip function always produce file in `/tmp` for instance), but this works and does not affect anything else.

### DIFF
--- a/pluginhelper.js
+++ b/pluginhelper.js
@@ -413,7 +413,7 @@ function publish() {
             
             zip();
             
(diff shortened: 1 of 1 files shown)


macmpi:
Similarly node modules directory get pushed also at submit, whereas plugin code repository does not seem supposed to contain such directory.
Should we move it away before submitting ?

PR #1311, opened 2017-08-21
Add sorting of Albums and Album Year

liberatorcnx:
Added variable 'albumsort' to define sorting and display of Albums.

Added variable 'albumdate'  to add Year to album displays and allow sub-sorting by year.

Added function 'dynamicSort' to allow sorting of a single object in array.

Added function 'dynamicSortMultiple' to allow sorting of multiple objects in an array.


volumio:
Interesting. Let me see if I got it:
- We need to set up UI options to sort albums by title or artist
- We need to set up UI to show album years, but if album year is enabled then albums are sorted by artists and then by year

If yes, the last is a bit cumbersome IMHO. In any case very interesting addition

liberatorcnx:
@volumio - Not quite right;

- We need to set up UI to sort albums by title OR artist

- We need to set up UI to show album years.  If set to 'true' then albums will display the year.  If albums are sorted by artist (with above selection) then it will sort by artist then by year.  If albums are sorted by title it will just add the year.

It makes sense to sort by artist then year if we can (if already sorted by artist).  There is no point in sorting by title then year.

Why is it cumbersome?  We can do it so we should :-)

volumio:
You're right, it makes sense. It will take a bit for me to do the UI as I am now reworking some other parts of the system. Do you mind if we wait a week or so?

liberatorcnx:
No problem to wait - it is a new feature so not critical.

volumio:
Sorry, never had the time to test it extensively. Did you close it for that or because you want to edit it?

primip:
what about this feature?

PR #1314, opened 2017-08-27
Do shutdown/reboot the systemd way

macmpi:
More conventional and more future-proof (i.e: debian-stretch)


volumio:
Was it tested just on rpi?

macmpi:
yes

volumio:
Works, thanks

PR #1316, opened 2017-08-30
Bluetooth support for mpd

geeks-r-us:
the bluetooth plugin need to be able to add audio outputs for the bluetooth speakers


volumio:
It is very interesting, but its a huge diff. Could you please do one with just your commits? 

To do this just fork master and then commit there just your commits. 
I think we could also refactor it to be a bit more "general purpose". Let me know

geeks-r-us:
Oh ok I got my mistake. Branch was based on master not on development.
Should I revert branch to base on development or just make a PR against master?

volumio:
Just against master, so at least I can see if we can make it multipurpose...

PR #1317, opened 2017-08-30
Bluetooth

geeks-r-us:
the bluetooth plugin need to be able to add audio outputs for the bluetooth speakers


geeks-r-us:
I found another use-case where a more generic way to edit mpd.conf would be helpful:
adding inputs like cdio_paranoia

volumio:
I think we can use this to add custom mpd config.
I would do something like this:
- If a plugin needs to write additional stuff to mpd.conf it will create a file in /tmp/mpdconf and place the additional conf there
- In mpd , when saving the conf, if /tmp/mpdconf exists, we integrate the content into mpd.conf 

What do you think? Do you think you can help doing this?

geeks-r-us:
Not sure if this leads to conflicts when this mechanism is used by multiple plugins

geeks-r-us:
I thought about it and a more flexible way would be a mechanism that allows every plugin to register call backs for each section in the mpd config and return it's own config parts

PR #1334, opened 2017-09-15
Logsubmit3

xipmix:
Another place for passwords to lurk was reported. 
Tested with 2.246 on rpi2. That turned up another oversight, that the backslashes in the sed  backreferences needed to be doubled in order to survive the various quotings as we construct the string that gets ```execSync()```ed.
To add to the fun, the ```<elided>``` string ends up being invisible when viewing the htmlified log in a browser window, but can be revealed by 'view source'. We wanted to hide that text anyway, so no further change needed.
Please test...

### DIFF
--- a/logsubmit.js
+++ b/logsubmit.js
@@ -49,12 +49,12 @@ for (var itemN in commandArray) {
 
 // remove sensitive information
 commandArray = [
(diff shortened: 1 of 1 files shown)


volumio:
Well done

PR #1346, opened 2017-09-26
add beaglebone and justboom capes

shawaj:
add beaglebone and justboom capes


volumio:
Let me know if there is progress on kernel side, and we will reopen in that case

shawaj:
This is a work in progress so can't we leave this open?

PR #1349, opened 2017-09-30
added callback mechanism which allows adding lines to mpd.conf

geeks-r-us:
as discussed in Bluetooth PR [ ](https://github.com/volumio/Volumio2/pull/1317)

for usage see: (https://github.com/geeks-r-us/volumio-plugins/tree/f/bluetooth/plugins/audio_interface/bluetooth_controller)

### DIFF
--- a/app/plugins/music_service/mpd/index.js
+++ b/app/plugins/music_service/mpd/index.js
@@ -32,7 +32,8 @@ function ControllerMpd(context) {
 	this.commandRouter = this.context.coreCommand;
 	this.logger = this.context.logger;
 	this.configManager = this.context.configManager;
-	this.config = new (require('v-conf'))();
+    this.config = new (require('v-conf'))();
+    this.registeredCallbacks = [];
 }
 
 // Public Methods ---------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


geeks-r-us:
Any comments on this ?

app/plugins/music_service/mpd/index.js
```diff
@@ -1007,7 +1008,13 @@ ControllerMpd.prototype.createMPDFile = function (callback) {
                 var conf12 = conf11.replace('"${ffmpeg}"', 'decoder { ' + os.EOL + 'plugin "ffmpeg"'  + os.EOL + 'enabled "yes"'  + os.EOL + 'analyzeduration "1000000000"'  + os.EOL + 'probesize "1000000000"' + os.EOL + '}' + os.EOL);
 			} else {
                 var conf12 = conf11.replace('"${ffmpeg}"', " ");
-			}
+            }
+            
+            for(var callback of self.registeredCallbacks)
+            {
+               var data = self.commandRouter.executeOnPlugin(callback[0], callback[1], callback[2]);
```
volumio: First problem is that this will work only with 1 callback. So please revise the for loop.

Second
Please don't use a straight array like this, as it's not really clear. Instead use an object. It will become like this:
var data = self.commandRouter.executeOnPlugin(callback.type, callback.plugin, callback.method, callback.data);
This will help developers understand what they need to put, and eventually also add a variable to pass to the function.

PR #1353, opened 2017-10-03
Favourites to plugins

volumio:


volumio:
@fanciulli adding a song to favourties causes the system to crash, removing a song from favourites causes a system to crash.
All favourites situations must be tested!

volumio:
@fanciulli it was an error merge and I manually reverted the two commits

PR #1355, opened 2017-10-05
Add to playlist from any music service plugin

sla89:
Hi,

I have fixed the problem that you cannot add songs/videos from spotify or youtube to a playlist managed by volumio.

A music service plugin that wants to support this too only has to implement a method called ```getTrackInfo``` that accepts the uri to be added. It must then return detailed information about the song or playlist.

I have added support for this in the youtube plugin and will create a  PR for it accordingly in the plugins repo (**see referenced PRs!**)

I had to adapt the code for removing elements from a playlist (method ```commonRemoveFromPlaylist``` line 602 -> the for loop): there was a check that compared the calling service name with the service name of the song. But in case of youtube or spotify the service name of an element is not mpd at it was expected and so the element was never removed. I think that checking the URI for removing the item should be sufficient enough anyway. 

Take a look at the plugin. It already seem to work for spotify.

### DIFF
--- a/app/playlistManager.js
+++ b/app/playlistManager.js
@@ -547,6 +547,33 @@ PlaylistManager.prototype.commonAddToPlaylist = function (folder, name, service,
 				});
 
 			});
+		} else {
+			// handle generally for all music service plugins
+			var result = self.commandRouter.executeOnPlugin('music_service', service, 'getTrackInfo', uri);
+			if (result) {
+				result.then(function (info) {
+					fs.readJson(filePath, function (err, data) {
+						if (err)
+							defer.resolve({ success: false });
+						else {
+							var output = data.concat(info);
+							fs.writeJson(filePath, output, function (err) {
+								if (err) {
+									defer.resolve({ success: false });
+								}
+								else {
+									defer.resolve(info);
+								}
+							})
+						}
+					});
+				}).fail(function (err) {
+					self.logger.error('Failed adding to playlist with error: ' + err);
+					defer.resolve({ success: false });
+				});
+			} else {
+				defer.resolve({ success: false });
+			}
 		}
 	});
 
@@ -571,7 +598,6 @@ PlaylistManager.prototype.commonRemoveFromPlaylist = function (folder, name, ser
 					defer.resolve({ success: false });
 				else {
 					var newData = [];
-
 					for (var i = 0; i < data.length; i++) {
 						if (data[i].uri !== uri) {
 							newData.push(data[i]);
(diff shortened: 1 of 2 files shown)


fanciulli:
Dear @sla89 ,

you said "I think that checking the URI for removing the item should be sufficient enough anyway". Unfortunately that's not true since there's no constraint for plugins to implement different uri schemas. This means that two plugins can manage the file:// uri, implementing two completely different behaviour.
That said your PR shall be updated accordingly.

sla89:
Dear @fanciulli ,

I think you are right. Anyway we have to fix volumio because when you remove something from a playlist volumio always passes ```service = 'mpd'``` into the function ```commonRemoveFromPlaylist``` which maybe is the root source of the problem. The API should be called with the original service name as it is stored in the playlist file. When you add something to a playlist the parameter ```service``` mentions the service (e.g. youtube) that wants to add something to a playlist which is then stored in the playlist file. But this information is ignored/gets lost when the playlist gets visualized to the user --> ```service```  is always ```mpd```.

I will take a look at it.

sla89:
Dear @fanciulli and @volumio 

I took a look into the problem and I figured out that the API is inconsistent in the WebSocket API. While ```addToPlaylist``` takes the service as parameter:

```json
   {
    "name":"my playlist",
    "service":"mpd",
    "uri":"USB/..."
   }
```

but ```removeFromPlaylist``` doesn't

 ```json
   {
    "name":"my playlist",
    "uri":"USB/..."
   }
```

Further in the plugin ```user_interface/websocket``` the service is always set to  ```mpd```which is absolutely wrong I think:  
https://github.com/volumio/Volumio2/blob/6edd104e4275096107d449ab3afb02abd749d87e/app/plugins/user_interface/websocket/index.js#L766-L769

Actually this is the problem why items from other music service plugins cannot be removed from a playlist anymore. Changing this would mean a breaking change in the API and will also affect the way the UI emits removeFromPlaylist but adapting that is not a big deal.

Shall I extend the PR that fixes that issue but has as side affect a breaking change of the API?

Looking forward to hear from you so that we can go on with this.

sla89:
@fanciulli  and @volumio I have removed the mentioned code above so that ```removeFromFavorites``` is not called with the hard coded service ```mpd```. I have also created another PR in the Volumio2-UI repo. The changed code in the UI simply passes now the service name when emitting ```removeFromFavorites```.

Further I am not sure what's going on here: 4d16dd85e3b7f47294eef4506336e15482910f03

sla89:
Done.

sla89:
@volumio no progress on that I guess, right?

volumio:
@sla89 again sorry for getting back very late with this. Could you fix the same issues as per my previous comment on the other pr?
Thanks...

sla89:
@volumio have you reviewed my PR and will accept it? Because @fanciulli still thinks there might be an issue.

Issue #188, opened 2016-01-25
RaspiDACRev.3 sound card support


jgrulich:
Please add the support for the raspidac3 overlay. If added to the config.txt, the card works, but the volume is set to 100% and can't be changed. The "Headphones" control works via amixer sset Headphones well. It will be fine if this control will be implemented in the Volumio too. Alsamixer already works with this.

apinner:
I can't read czech, but presumably this should just be a case of adding a new line in dacs.json?

http://www.grulich.eu/files/RaspiDACv3.pdf

volumio:
Yep!

apinner:
Added in https://github.com/volumio/Volumio2/commit/c08011ef497ef96464e279049168f5299424cac8

volumio:
Thanks!

jgrulich:
There are two controls for this card. There is "Digital" for DAC volume control and "Headphones" for headphones amplifier volume control. If there is no option for second knob, than let's make two card versions, one with 100% "Headphones" and regulated "Digital" and second one swapped.

apinner:
`{"id":"raspidac3dac","name":"RaspiDACv3 DAC","overlay":"raspidac3","alsanum":"1","mixer":"Digital","modules":"","script":""},
{"id":"raspidac3hp","name":"RaspiDACv3 Headphones","overlay":"raspidac3","alsanum":"1","mixer":"Headphones","modules":"","script":""}`

???

jgrulich:
Yes, somehow it works. Exactly it regulates the given volume, but doesn't set the 100% for the second one. It lives it as it was before. Means that when I use the "DAC" and set the volume to 10% and than switch to the "Headphones" than I'm not able to increase the DAC to 100% until I use the alsamixer via the SSH.

jgrulich:
I've tried to use this:

{"id":"raspidac3dac","name":"RaspiDACv3 DAC","overlay":"raspidac3","alsanum":"1","mixer":"Digital","modules":"","script":"amixer -c 1 sset Headphones 40"},
    {"id":"raspidac3hp","name":"RaspiDACv3 Headphones","overlay":"raspidac3","alsanum":"1","mixer":"Headphones","modules":"","script":"amixer -c 1 sset Digital 200"}

but it doesn't work. Another way how to run the init script?

apinner:
@volumio any thoughts on this one?

jgrulich:
[...]

PR #1357, opened 2017-10-07
Localization support for plugins

sla89:
Currently the plugins have to implement localization support by their own. The ```CoreCommandRouter ``` called the method ```loadI18NStrings``` on each plugin if the method exists. But this was not documented at all so the method is not used so far.

I have replaced this method by a new method called ```getI18nFile```. The idea is that each plugin returns the i18n filename that should be loaded into volumio on startup.  Then volumio reads that file and merges the locales into the variable ```i18nStrings```. Now the plugins can call ```self.commandRouter.getI18nString('MY_LOCALE_KEY')``` and they get the localized string for it. 

With this feature we can now provide localized toast messages in the spotify, qobuz and youtube plugin.

When you accept the PR I will update the docs accordingly where we also have to setup some guidelines so that we have no merge issues. My code already takes aware of duplicated keys and only adds new keys to ```i18nStrings```.

The implementation of ```getI18nFile``` could look like this and could be part of the example plugin too:

```javascript
Youtube.prototype.getI18nFile = function (lang_code) {
  var supportedLanguages = ['en', 'it', 'de'];

  if (supportedLanguages.includes(lang_code)) {
    return path.join(__dirname, 'i18n', 'strings_' + lang_code + '.json');
  }

  // return default i18n file
  return path.join(__dirname, 'i18n', 'strings_en.json')
}
```

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -12,76 +12,76 @@ var vconf = require('v-conf');
 module.exports = CoreCommandRouter;
 function CoreCommandRouter(server) {
 
-	var logfile = '/var/log/volumio.log';
+  var logfile = '/var/log/volumio.log';
 
-	fs.ensureFileSync(logfile);
-	fs.watchFile(logfile, function () {
-		fs.stat(logfile, function (err, stats) {
-			if (stats.size > 15728640) {
-				var now = new Date();
-				console.log('******** LOG FILE REACHED 15MB IN SIZE, CLEANING IT ********');
-				fs.writeFile(logfile, '------------------- Log Cleaned at '+ now + ' -------------------', function(){
-					console.log('******** LOG FILE SUCCESSFULLY CLEANED ********');
-				})
-			}
-		});
-	});
-	this.logger = new (winston.Logger)({
-		transports: [
-			new (winston.transports.Console)(),
-			new (winston.transports.File)({
-				filename: logfile,
-				json: false
-			})
-		]
-	});
+  fs.ensureFileSync(logfile);
+  fs.watchFile(logfile, function () {
+    fs.stat(logfile, function (err, stats) {
+      if (stats.size > 15728640) {
+        var now = new Date();
+        console.log('******** LOG FILE REACHED 15MB IN SIZE, CLEANING IT ********');
+        fs.writeFile(logfile, '------------------- Log Cleaned at ' + now + ' -------------------', function () {
+          console.log('******** LOG FILE SUCCESSFULLY CLEANED ********');
+        })
+      }
+    });
+  });
+  this.logger = new (winston.Logger)({
+    transports: [
+      new (winston.transports.Console)(),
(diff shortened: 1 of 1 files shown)


volumio:
@sla89 can you please solve the conflicts? Also, can you contact me directly at info at volumio dot org ? 
THANKS!

sla89:
@volumio I have resolved them and I have sent you a mail.

volumio:
@sla89 sorry if I get back so late with this. 
There is a problem with this PR: the diff is huge because of your different linting. If you could redo it with the same linting I would better appreciate its content and it could be safely merged. While you're there it would be great if you could solve the conflict... thanks

volumio:
@sla89 Just thought now that we have an issue with this:
this works when volumio starts, but what happens if we just install and then enable the plugin?

gvolt:
I can confirm the problem volumio pointed out on 6 Feb.

Workaround is to call the method "self.commandRouter.loadI18nStrings()" on every start of a plugin (onStart method), see PR #188.

sla89:
Sorry for the late response. I don't think that this workaround is a good idea, because you will always load all localizations again and not just the one of the plugin.

Further I forgot to cleanup the store for all the terms.

I guess a better approach would be to extend my PR by moving the code for loading the terms of a plugin into a method and call this method in `CoreCommandRouter.prototype.enablePlugin` and cleanup things in `CoreCommandRouter.prototype.disablePlugin`.

What do you think about that? I guess that should fix all the problems. I would do the loading of the terms after the plugin manager successfully loaded/disabled the plugin.

gvolt:
Yes, I fully agree. It is not a nice solution at all, but I did not find another one ;-)

I am all for handling it in `CoreCommandRouter.prototype.enablePlugin` / `...disablePlugin`.

While we are at it. I extended your suggested implementation of `getI18nFile` in order to automatically determine the supported languages and not have to adjust the plugin's code only because a new language has been added:
```
<plugin_name>.prototype.getI18nFile = function(lang_code) {
    var supportedLanguages = fs.readdirSync(__dirname + "/i18n");

    // determine supported languages
    for (var i = 0; i < supportedLanguages.length; ++i) {
        if (supportedLanguages[i].startsWith("strings_") && supportedLanguages[i].endsWith(".json")) {
            supportedLanguages[i] = supportedLanguages[i].replace("strings_", "").replace(".json", "");
        } else {
            supportedLanguages.splice(i, 1);
        }
    }
    if (supportedLanguages.includes(lang_code)) {
        return path.join(__dirname, "i18n", "strings_" + lang_code + ".json");
    }

    // return default i18n file
    return path.join(__dirname, "i18n", "strings_en.json");
};
```
Let me know what you think about it.

gvolt:
@sla89 Did you have the opportunity to have a look at this issue?

PR #1373, opened 2017-10-25
Promises4on volumio shutdown&reboot

fideleinratio:
Currently the onShutdown and onReboot functions are launched, but nothing is in place to be sure that their routines (if used) are finished before the shutdown or reboot respectively. 

simply adding the ability (and requirement) to use promises - shutdown / restart occurs after all promises returned from the plugins are resolved or rejected. 
reused code by @macmpi  within plugin manager to create fake promises for plugins with either onVolumioShutdown or onVolumioReboot functions but which do not return a promise. 

tested ok for compliant and non-compliant plugins on pi3

@macmpi  and @volumio  comments or issues please?? 

also, hope this is correct process. not a programmer, first time participating. pls let me know if not


fideleinratio:
To give more context for why this is proposed: 
one of the community's plugins: the Audiophonics RaspDAC by @JedS has a onVolumioShutdown routine that, on my side at least, does not work correctly. there is a GPIO that is turned on and then off after 1 second. however before the one second is timed-out, the system goes into shutdown so the plugin never works as intended. 

if this PR is implemented, the above plugin, and any others that might have similar problems can be easily updated to ensure completion before shutdown just by returning a promise and resolving it on completion.

@macmpi, @volumio and @JedS any comments, questions, or feedback welcome.

macmpi:
@fideleinratio thanks for your note and reference.
Indeed, when I added promise check I considered adding same for onVolumioShutdown & al, but did not as I could not test it.
It's probably fine doing it, though I did not fully read/test the proposed PR: maybe @JedS can confirm as his plugin might be the only one using this hook currently.

fideleinratio:
@macmpi, thanks for the feedback
@JedS will create a fork of the raspDac plugin using a promise for onVolumioShutdown() in case you have the opportunity to test it with this PR.
for info have implemented this locally and it works as expected.

JedS:
@fideleinratio sounds pretty good to me. I'll test it asap as the PR will be merged.

volumio:
Very very useful, thanks!

PR #1381, opened 2017-11-05
fix url to correct github awesome-raspberry-pi list

basnijholt:
https://github.com/thibmaek/awesome-raspberry-pi is the correct repo


volumio:
Thanks, didn't notice it was forked

PR #1420, opened 2017-12-18
Allow adding a playlist and start playing it from commandline.

ways:


volumio:
I would suggest sticking to the format of volumio command line client:
https://github.com/volumio/Volumio2/blob/master/app/plugins/system_controller/volumio_command_line_client/volumio.sh

What you need to do is add the addplay there and edit the API client accordingly:
https://github.com/volumio/Volumio2/blob/master/app/plugins/user_interface/rest_api/index.js

volumio:
@ways what do you think of my comment?

ways:
You're correct, but I don't have time to rewrite this at the moment, and this solution works for me. Sorry. =)

PR #1428, opened 2017-12-22
Make 'volumio vstop' work

### LINKED ISSUE #1427: CLI deamon stop command
https://github.com/volumio/Volumio2/blob/f49ac565fb0c463f65dbc5166c1f779059c6f1a1/app/plugins/system_controller/volumio_command_line_client/volumio.sh#L62

This script contains `vstop()` but L138 and L142 are calling just `stop`.

xipmix:
While we're there, name the other commands controlling the volumio service consistently.
I couldn't find any internal commands using these but perhaps this is an API break.

Fixes #1427. Not tested beyond making sure 'volumio vstop' doesn't fail with 'command not found'.


volumio:
Makes lots of sense

Issue #1434, opened 2017-12-26
Sleep timer should start from 00:00 if it's not already running


teq0:
If you set a sleep timer for, say, 10 minutes, and go back to the sleep timer dialog after 12 minutes, the value for the sleep timer will say 23 hours 58 minutes (or so). The sleep timer modal reads the current value when it loads, which is good, however it still calculates the time from now until the expiry time, even if that has expired.

There's a secondary effect of this, which is that setting the sleep timer to 00:00 doesn't do anything (maybe the UI should treat this as disabling the timer rather than saving a setting of 00:00), but if you go back in a few minutes later it thinks the timer is running and has 23 hours and fifty something minutes to go.

The initial values in the UI should be 00:00 if the timer is not currently active. I've submitted this to the back-end as it feels cleaner if the back-end (socket interface) reports the correct thing.

teq0:
Fixed in https://github.com/volumio/Volumio2/pull/1435

PR #1435, opened 2017-12-26
Sleep timer disables once triggered, returns 00:00 via getSleep when …

teq0:
Fix for https://github.com/volumio/Volumio2/issues/1434

Set sleep_enabled flag to false when sleep timer is triggered, plus return 00:00 as the sleep time unless the sleep timer is actually active.

Indentation is a bit weird in some of these files, I didn't attempt to fix it.


volumio:
Thanks!

Issue #1458, opened 2018-01-16
Non-functional dummy-albums in Genre and Artist views


andreas-henning:
I will be providing a PR for this soon. This just provides a description so I have something to reference from the commit.

In the Genres and Artists views, orphaned tracks (i.e. tracks that don't belong to an album) are grouped into a dummy-album. Clicking on the dummy-album typically lists all of the album-artist's orphaned tracks. It is currently possible for the dummy-album to list no orphaned tracks at all, making it non-functional.

Problem:

This issue arises if the last (or only) track returned by:
```js
ControllerMpd.prototype.listArtist(curUri,index,previous,uriBegin)
	self.clientMpd.sendCommand(cmd(findartist, []), function (err, msg)
```
is an orphaned track. The subsequent call to:
```js
self.searchFor(lines, i + 1, 'Album:')
```
is expected to either find the track's 'Album:' entry, or, if it's an orphaned track,
find the the next track's "file:" entry signaling to the procedure that the "Album:" entry is missing.

If the orphaned track is the last track in the list, the procedure will not encounter a subsequent "file:" entry, which causes the function to exit without a return value. The lhs value at the call site (a variable intended to contain the album name) becomes undefined. This of course causes other issues, like the inability for volumio to find any tracks belonging to an album with that name.

volumio:
Fixed with https://github.com/volumio/Volumio2/commit/5d40d4bd58a14e01628dc89d348c95cd3018d1d5

Issue #1453, opened 2018-01-12
Limit or eliminate the creation of dummy-albums for orphaned tracks


andreas-henning:
Dummy albums are created to serve as a container for orphaned tracks (i.e. tracks without an album).

I propose that dummy-albums should be removed from Artists, Genres and Albums views. In the Artists and Genres views, orphaned tracks will remain accessible through the tracks-list. Orphaned tracks would no longer be accessible from the Albums view, which might pose a problem for some.

On the other hand, I can't imagine this being useful in a practical sense. If we want to play an artist's tracks (including orphaned tracks), we can do that from the 'Artists' view. If we want to play tracks of a certain genre (including orphaned tracks), we can do that from the 'Genres' view. What exactly is the use case where somebody would want to play only an artist's orphaned tracks?

For people with music collections with many orphaned tracks, removing dummy-albums will dramatically shorten the list of albums in the 'Albums' view, making it much easier to find real albums.

For people who have a clean music collection, where every track is assigned to an album, this proposal would change nothing.

Ideas? Feedback?

UPDATE:
Previously I mentioned that dummy-albums can occasionally have no tracks, making them non-functional. I've since created a separate issue for that, which is now here:
https://github.com/volumio/Volumio2/issues/1458

xipmix:
One use I can think of for displaying orphaned tracks is so they could be cleaned up, ie assigned to some parent by moving to the subdirectory in the music collection the user deems appropriate. I think flagging them as orphans in the 'Album' view, while making them available as normal under 'Artist' or 'Genre', as you propose, is a reasonable approach. 

Does it make any sense to collate all orphans, from whatever artist, in one 'orphans' dummy album in the 'Album' view? That also tidies up the Album listng.

andreas-henning:
@xipmix 

> I  think flagging them as orphans in the 'Album' view, while making them available as normal under 'Artist' or 'Genre', as you propose, is a reasonable approach.

Did I understand you correctly:

It would be okay to remove dummy-albums from the 'Artists' view (as is already the case in the 'Genres' view), because in both of those views orphaned tracks are still accessible in the list of tracks.

I would certainly agree with that. I also like your idea to collate orphans in the 'Albums' view. IMHO that would nicely clean up collections with many orphans without sacrificing functionality.

I could get behind all of that.

[...]

Issue #1441, opened 2017-12-31
if id3tag %track% are empty, enabling "show track number" result in UNDEFINED


Bescio:
When enabling "Show Track Numbers" in the Volumio options, tracks without the ID3 %track% populated, show a "UNDEFINED" value before the title. I'm using latest version of Volumio installed from scratch. 

Please look at the screnshoots:

![undefined](https://user-images.githubusercontent.com/34973944/34458630-4519cd36-edda-11e7-8695-7a77dd646353.JPG)
![undefined1](https://user-images.githubusercontent.com/34973944/34458631-484e86ae-edda-11e7-9ae4-0c38e73e3595.JPG)
![undefined2](https://user-images.githubusercontent.com/34973944/34458632-4b161ab4-edda-11e7-8653-8c5c9495602f.JPG)

Thank you.

volumio:
Fixed with https://github.com/volumio/Volumio2/commit/5d40d4bd58a14e01628dc89d348c95cd3018d1d5

PR #1460, opened 2018-01-16
Improved handling of orphaned tracks

andreas-henning:
Fixes these issues:

volumio/Volumio2/[#1458](https://github.com/volumio/Volumio2/issues/1458)
volumio/Volumio2/[#1453](https://github.com/volumio/Volumio2/issues/1453)
volumio/Volumio2/[#1441](https://github.com/volumio/Volumio2/issues/1441)  [Likely fix, not 100% sure]

For people who's collections have no orphaned tracks (tracks without an album), these changes should make no difference.

@volumio 

Hey Michealangelo

I know you're super busy. If you can find the time I'd appreciate you taking a quick look at my two PRs and providing feedback. There is a bit more I'd like to contribute (sorting and performance related), but that all builds on top of the two PRs I've already made, so I'd first need to know if they actually make any sense and can be accepted into the master repo before I continue.

### DIFF
--- a/app/plugins/music_service/mpd/index.js
+++ b/app/plugins/music_service/mpd/index.js
@@ -2979,11 +2979,17 @@ ControllerMpd.prototype.listAlbums = function (ui) {
 						if (!artistName) {
                             artistName = self.searchFor	(lines, i + 1, 'Artist:');
 						}
-					//********Check if album and artist combination is already found and exists in 'albumsfound' array (Allows for duplicate album names)
-						if (albumsfound.indexOf(albumName + artistName) <0 ) { // Album/Artist is not in 'albumsfound' array
-							albumsfound.push(albumName + artistName);
-							var codedArtistName = encodeURIComponent(artistName);
-              				var codedAlbumName = encodeURIComponent(albumName);
+                        // This causes all orphaned tracks (tracks without an album) in the Albums view to be
+                        //  grouped into a single dummy-album, rather than creating one such dummy-album per artist.
+                        var albumId = albumName + artistName;
+                        if (!albumName) {
+                            albumId = '';
+                            albumName = '';
+                            artistName = '*';
+                        }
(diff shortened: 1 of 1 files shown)


andreas-henning:
Hey xipmix, thank you very much for your feedback!

> What are you meaning to test with if (!albumName) - that albumName is undefined, or just an empty string?

Both actually ;-) While technically different (undefined should probably imply "error" / empty should probably imply "missing metadata"), I though it best to make no distinction between them behaviorally. Volumio can't know what album either track belongs to, so both are treated like orphans, meaning both undefined and empty tracks are assigned to the dummy-album. This signifies to the user that those tracks may require some cleanup.

Ideally, Volumio's functions would be more contractual. Then I'd say "undefined" should throw an exception signaling that its contract was broken. It just doesn't seem like this would be in the spirit of how Volumio currently does things.

What is your take?

> Might be worth explaining in the commit that the coded<foo>Name variables are superfluous here, or else make use of them as is done in the listArtists function.

I suck at writing good commit messages :-/

The thing is, figuring out what unused code was initially intended to do is often very hard. It's just left over for whatever reason, and finding and documenting how we got to that point where it's no longer required is time consuming. I'd prefer to just ignore that.

Of course I could just add to the commit message that I "removed unused code", but that is so trivial I'm not sure that would really be helping anybody. Am I misunderstanding you or wrong?

xipmix:
Thanks for all the explanations. I don't think there's much point changing your PR untill @volumio weighs in. It would nice to handle undefined and empty string cases more explicitly but it can wait.
Re the commit message, don't bother changing it. If it was me, I would probably have said 'dropped unused variables' or some such.

andreas-henning:
Ok, appreciate the feedback!

volumio:
From my tests, that's actually quite good! 
Thanks, appreciate your contributions!

PR #1471, opened 2018-02-05
Updated spanish translation

webodan:
Hello,

I found the spanish translation to be missing some strings as well as others would read strangely.
I'm creating this pull request and attaching my own updated version to make up for that. 
I couldn't get it to work because the program seems to expect the new strings into some other part of code. But the translation now has the same amount of strings in the same positions as the English one.

[strings_es.zip](https://github.com/volumio/Volumio2/files/1695346/strings_es.zip)


volumio:
Thanks! The missing translations are for the UI, please see how to do it here:
https://volumio.org/forum/contribute-translating-volumio-your-language-t4289.html

Gracias!

PR #1455, opened 2018-01-14
Properly escape double quotes when constructing mpd command strings

xipmix:
Issues #1454 and #1430 raise a problem with music collection files
or album names containing double quotes. This arises from the idiom
used to construct commands sent to mpd -
```
  command = "action \"" + some_datum + "\"";
```
In this series I try to fix this by adding escaping backslashes
to any 'some_datum' string when putting the command string together.

I reproduced both the issues above with a small test music collection:
```
test/
     U2/
        File_"with_quotes"_.flac
     You "Two"/
               File_with_no_quotes.flac
 ```
The current code fails in the following ways:
```
Browse->MusicLibrary->NAS->test
  both folders appear,
  but when visiting test->You "Two"/ no tracks are listed

Browse->MusicLibrary->NAS->test->U2
  the File_"with_quotes"_.flac track is listed but cannot be queued or played
```
I have confirmed that after my changes these problems go away.
I also tried a file with double quotes in the name via UPNP,
and web radios, though not with " in any URLs. These sources
both worked without problems.

Still, quite a few changes here, best to test independently.

Issue #1454, opened 2018-01-13
Can't play file with double quote character in name


kaare:
I started to rip my CD box with the complete works of Verdi. The music files (ogg) are placed on a NAS, , and I make playlists in Volumio.

I didn't notice any problems until I reached Rigoletto, where all files except the Sinfonia have '"' in the names. And he playlist only contained the one file without.

I can't play the file either, so it's not only a playlist problem.

I digged a little into this, and this seems to be the relevant part of the log

```
2018-01-13T09:27:05.584Z - info: CoreCommandRouter::volumioAddQueueItems
2018-01-13T09:27:05.585Z - info: [1515835625585] CoreStateMachine::addQueueItems
2018-01-13T09:27:05.586Z - info: [1515835625586] CorePlayQueue::addQueueItems
2018-01-13T09:27:05.587Z - info: Adding Item to queue: music-library/NAS/NAS/Music/Opera/Verdi/Verdi; Shirley Verrett, Piero Cappuccilli, Plácido Domingo, Nicolaï Ghiaurov, Coro e Orchestra del Teatro alla Scala, Claudio Abbado - Macbeth (disc 2)/01 - Giuseppe Verdi - Macbeth: Atto III. "Tre volte miagola".ogg
2018-01-13T09:27:05.588Z - info: Exploding uri music-library/NAS/NAS/Music/Opera/Verdi/Verdi; Shirley Verrett, Piero Cappuccilli, Plácido Domingo, Nicolaï Ghiaurov, Coro e Orchestra del Teatro alla Scala, Claudio Abbado - Macbeth (disc 2)/01 - Giuseppe Verdi - Macbeth: Atto III. "Tre volte miagola".ogg in service mpd
2018-01-13T09:27:05.599Z - info: [1515835625599] CorePlayQueue::saveQueue
2018-01-13T09:27:05.602Z - info: CoreCommandRouter::volumioPushQueue
2018-01-13T09:27:05.603Z - info: [1515835625602] interfaceApi::pushQueue
2018-01-13T09:27:05.604Z - info: [1515835625603] InterfaceWebUI::pushQueue
2018-01-13T09:27:05.608Z - info: [1515835625607] CoreStateMachine::updateTrackBlock
2018-01-13T09:27:05.609Z - info: [1515835625608] CorePlayQueue::getTrackBlock

... and then nothing. If I play the next file, one without '"' in the name, the log is much, much longer, first lines are 

[...]

Issue #1430, opened 2017-12-23
Files with double quotes in tags are not displayed


kaito3desuyo:
Hello.
I am using RasPi 3B and Volumio 2.348 with PCM5102A I2S DAC.

When displaying on album basis on the Web UI, an event occurred in which a file containing double quotation marks could not be displayed.

Are there any good ideas?
Sorry for machine translation.
Thank you.

xipmix:
Best idea would seem to be to urlencode links to album tracks, see volumio/Volumio2-UI#334

kaito3desuyo:
Thank you for your help.
Specifically, how do you think to solve it?

xipmix:
I think this will need logs from your system to solve.

1. Reboot the system
2. select album view, find the view where the album with double quotes is not displayed
3. In the web browser URL bar, enter ```http://<volumio ip addr>/dev```
4. Click on the 'send logs' button near the bottom of the page.
  It will prompt you for a comment, please enter something like 'douvle quotes issue 1430'.
5. It will upload and return a URL. Please paste URL that here.
 The logs should be cleansed of sensitive material; pease open an issue if you find things in the uploaded log that you think should be filtered out.

kaito3desuyo:
okay, i uploaded log to [http://logs.volumio.org/volumio/1wSMkdI.html].

xipmix:
Thanks for your log, I think it lead me to the problem, which I've proposed a fix for.

PR #1482, opened 2018-02-18
Properly escape double quotes when constructing mpd command strings - respin

xipmix:
This replaces PR #1455. See also Issues #1454 and #1430

### DIFF
--- a/app/plugins/music_service/mpd/index.js
+++ b/app/plugins/music_service/mpd/index.js
@@ -3313,10 +3313,10 @@ ControllerMpd.prototype.listArtist = function (curUri,index,previous,uriBegin) {
 
     self.mpdReady
     .then(function() {
-        var artist=decodeURIComponent(splitted[index]);
+		var artist=decodeURIComponent(splitted[index]);
 		var VA = 0;
-        var cmd = libMpd.cmd;
-        var safeArtist = artist.replace(/"/g,'\\"');
+		var cmd = libMpd.cmd;
+		var safeArtist = artist.replace(/"/g,'\\"');
 
 		if (uriBegin === 'genres://')  {
 			var genre = decodeURIComponent(splitted[2]);
--- a/app/plugins/music_service/mpd/index.js
+++ b/app/plugins/music_service/mpd/index.js
@@ -1191,8 +1191,9 @@ ControllerMpd.prototype.lsInfo = function (uri) {
 		prev = sections.slice(0, sections.length - 1).join('/');
 
 		folderToList = sections.slice(1).join('/');
+		var safeFolderToList = folderToList.replace(/"/g,'\\"');
 
-		command += ' "' + folderToList + '"';
+		command += ' "' + safeFolderToList + '"';
 
 	}
 
@@ -2471,7 +2472,8 @@ ControllerMpd.prototype.scanFolder=function(uri)
             if (sections.length > 1) {
                 folderToList = sections.slice(2).join('/');
 
-                command += ' "' + folderToList + '"';
+                var safeFolderToList = folderToList.replace(/"/g,'\\"');
+                command += ' "' + safeFolderToList + '"';
 
             }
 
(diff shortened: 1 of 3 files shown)


xipmix:
I doubt I have any credibility with the community. 

I appreciate you are short of review time. Can you briefly outline what your test process would be (so others can reproduce it)?
The tests I did are described in #1482, if people can do something like that would that be sufficient?

Hmm. What if we add something to /usr/local/bin/volumio.sh that allows someone to pull a PR into their volumio, and then revert the change when they are done? The PR submitters would need to give some notes on what they want people to check.

volumio:
Credibility is something you build with your work, and you have plenty of credit to be credible.

I usually play my tracks for a whole day and try to crash the thing.

Imho it would be wiser to give the exact commands as of now, and then ask to submit result. Once done its sufficient to delete user data. 

But we could set up a way to test PRs as you ask, it would be super useful

xipmix:
> Imho it would be wiser to give the exact commands as of now, and then ask to submit result. Once done its sufficient to delete user data.

had trouble parsing that - I take it you mean 'give the exact commands for people to do the test with', and then they should put comments on the result in the PR ? A test script that produces a pasteable log perhaps?

I'll look into the 'prtest' idea a bit more.

volumio:
I would go with just a forum post with the instructions and list of command to test

chsims1:
I'm very happy to do tests, as long as it's spelled out exactly how and what you want doing.  As for credibility, I think you are doing a wonderful job xipmix ;)

I think the forum post idea is the simplest way forward to involve more testers .... if you don't want to post on forum, then put the info here & I'll post it for you.

xipmix:
@chsims1 Thanks for the kind words. How about trying this yourself and if it works others could try. I'm happy for you to post to the forum. We probably only need a few positive results.

What the patch is trying to do is improve the handling of albums that have double quote marks (") in their names. If you have an album like that in your collection it will probably not show up when browsing albums; this patch intends to fix that.

To test the patch,
1. Attached is a small shell script [test.txt](https://github.com/volumio/Volumio2/files/1739581/test.txt)  . Save it and upload into your volumio
```
$ scp test.txt volumo@volumio:test.sh
```
2. Run the script to apply the patch, and restart volumio
```
volumio@volumio$ sh test.sh
Patch applied successfully. Please restart volumio
To undo the patch:  cd /volumio; git reset --hard
volumio@volumio$ sudo systemctl restart volumio
```
3. Browse your albums to check if the missing albums appear.

To report results,
1. Please give your system version
```
volumo@volumio$ cat /etc/os-release
```
2. If the patch failed to apply, please report what was printed out.
3. Please report any changes you noticed when browsing your albums, particularly if items with " in the name appeared.

chsims1:
OK, might not be able to try until the weekend ... will see how things go.

xipmix:
Ok thanks for taking the time to try this. It doesn't look like a feasible approach - it seems it would require rebasing the patches to the last revision that went into the release you are running. That could be done, but then we are no longer testing the thing @volumio needs us to test, that merging into the current master (e89dda7f2d3f5b8120b2923126c4acde2f9d395e Fix plugin uninstall, update and unzip) will work and not break anything.

chsims1:
OTA update to 2.368 and patch is now applied.

I have always removed double quotes from my music files as a matter of course, so for this test I introduced them into one track file name and the folder name holding the track.

1. Before patch.

Track filename with quotes: I can see the track but I can't play it.
Folder name with quotes: I can see all the tracks but can't play any of them both by artist and album view.  If I browse the 'Music Library' and look under NAS etc, the I can see the folder, but nothing in the folder.

2. After patch.

Everything works fine :D.  I can see and play everything in whatever browsing view.

No difference in shown numbers of artists, albums and tracks before or after patch.

xipmix:
Good news, thanks for the test.

kaito3desuyo:
@xipmix Thank you for creating the patch. 
Patch worked fine on volumio 2.368.

volumio:
Thanks @xipmix and @kaito3desuyo @chsims1 !

Merging

Issue #1484, opened 2018-02-18
Webradio search woes


macmpi:
It seems Web radio search/parsing changed sometimes back around last December (maybe linked to dirble vs shoutcast changes ?).
Since then I've noticed some search issues, like duplicated results (with some failing), missing ones, etc...

Here is an example with easy steps to replicate:
- search webradio for "audiophile"
- result lists several radios, including **Audiophile Jazz** twice (first one failing to play)
- saved those two in Webradio Favorites
- cat the radio-favourites file

here it is:
```
volumio@volumio:~$ cat /data/favourites/radio-favourites
[{"service":"webradio","uri":"http://8.38.78.173:8276/\n","title":"Audiophile Jazz","icon":"fa-microphone"},{"service":"webradio","uri":"http://8.38.78.173:8210/stream","title":"Audiophile Jazz","icon":"fa-microphone"}]
```
note the suspect **\n** character in the first URL, which is not playing. Second one is fine.

xipmix:
I can confirm the behaviour. The ```\n``` characters are coming from the URIs in the webradio database. I traced through the resolution of the 'Audiophile Jazz' station. In plugins/music_service/webradio/index.js, ```ControllerWebradio.prototype.listTop500Radios``` does roughly this:
 * a query to ```http://api.shoutcast.com/legacy/Top500``` returns an xml file
 * with a ```<station>``` element that has the ```id=1459011```
 * constructs a uri pointing to that id on ```http://yp.shoutcast.com```

Somehow later that uri is visited. Doing that manually returns this m3u file
```
#EXTM3U
#EXTINF:-1,(#1 - 282/1000) Audiophile Jazz
http://8.38.78.173:8210/stream
```
The last line has a ```\n``` at the end. Whatever is parsing that file to extract the uri isn't dealing with that. I haven't figured out where that happens.

xipmix:
I found that I'd misunderstood some things here.
When searching as above, ControllerWebradio.prototype.search queries the dirble service. It asks for json to be returned, which should cause no issues with parsing. However the json that comes back has the bogus uri embedded in the 'streams' array:
```
{ id: 9471,
  name: 'Audiophile Jazz',
  country: 'GR',
  image: {
    url: 'https://img.dirble.com/station/9471/audiophile.jpg',
    thumb: {
      url: 'https://img.dirble.com/station/9471/thumb_audiophile.jpg'
    }
  },
  slug: 'audiophile-jazz',
[...]

PR #1493, opened 2018-02-28
Ensure clean uris from Dirble

xipmix:
Some uris in the dirble search results contain end of line characters,
which break things downstream in volumio. Stripping them away makes things
work again.

This addresses #1484. Lightly tested against 2.362 (rpi2).


volumio:
Well catched!

PR #1503, opened 2018-03-08
Add Fe-Pi Audio Cards

Fe-Pi:


volumio:
Hi, can you confirm that works even without 
dtoverlay=i2s-mmap

added?

Fe-Pi:
I have confirmed it works without dtoverlay=i2s-mmap in the config.txt

volumio:
Thanks, merging

PR #1514, opened 2018-03-13
Utilise `currentSeek` to track Seek in consume mode

### LINKED ISSUE #1510: Playback timer on web interface resets to 0:00 when changing volume
Hello,
On the latest stable release of Volumio2, with a generic pcm2704 USB DAC used for playback, as i'm playing an audio track, if I raise or lower the volume of the currently playing track, the timer resets to 0:01 and continues counting. The timer gets the correct playback timing again if I click on "Library" and then back to "Playback". I haven't changed anything else on the playback settings except for the output device.

This is the output I get in volumio.log as I change the volume:

`2018-03-13T12:27:12.799Z - info: VolumeController::SetAlsaVolume55
2018-03-13T12:27:13.049Z - info: VolumeController::Volume 55
2018-03-13T12:27:13.071Z - info: CoreStateMachine::pushState
2018-03-13T12:27:13.085Z - info: CoreStateMachine::getState
2018-03-13T12:27:13.095Z - info: CoreCommandRouter::volumioPushState
2018-03-13T12:27:13.100Z - info: CoreCommandRouter::executeOnPlugin: volumiodiscovery , saveDeviceInfo
2018-03-13T12:27:13.118Z - info: interfaceApi::pushState`

ashthespy:
When in consume mode, while the UI displays the timer correctly, state information is rest to the original `consume.state` when `pushState` is called from the state machine's internal functions, such as `updateVolume` for example.

This is a quick and dirty fix for that, by piggy backing the already present `currentSeek` value.
I realise that isn't ideal, but ideas welcome, will implement! :-)

Fixes #1510 
and solves the issue from my old question on the forum https://volumio.org/forum/pushing-playback-state-info-volumio-t7641-10.html back from volspotconnect days :-)

### DIFF
--- a/app/statemachine.js
+++ b/app/statemachine.js
@@ -423,7 +423,7 @@ CoreStateMachine.prototype.increasePlaybackTimer = function () {
 
 	var now=Date.now();
 	this.currentSeek+=(now-this.playbackStart);
-	this.isConsume?  this.consumeState.seek = this.currentSeek : null;
+	this.isVolatile?  this.volatileState.seek = this.currentSeek : null;
 	if(this.runPlaybackTimer==true)
 	{
 		this.playbackStart=Date.now();
--- a/app/statemachine.js
+++ b/app/statemachine.js
@@ -417,13 +417,13 @@ CoreStateMachine.prototype.getNextIndex = function () {
 };
 
 
-// Stop playback timer
+// Increase playback timer
 CoreStateMachine.prototype.increasePlaybackTimer = function () {
 	var self=this;
 
 	var now=Date.now();
 	this.currentSeek+=(now-this.playbackStart);
-
+	this.isConsume?  this.consumeState.seek = this.currentSeek : null;
 	if(this.runPlaybackTimer==true)
 	{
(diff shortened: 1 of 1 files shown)


ashthespy:
Actually, I don't think this fixes the problem -- would love some inputs from the statemachine experts!

volumio:
@ashthespy this is a regression happened when we did not retrieve the state back from state machine at every volume change.
IMHO your solution might work, but we must then ping back the statemachine value at every volume change.

ashthespy:
@volumio  my commit is essentially nonsense. I got confused b/w volatile and consume to begin with. I shall close this PR if I don't figure something out in the next few hours ;-)

But I agree, the statemachine is quite a beast especially for a newcomer just looking at the code/comments!

ashthespy:
FWIW: this currently works (not entirely) - but only if you perform a `seek` once from the UI, which seems to triggering `startPlaybackTimer`.

ashthespy:
Closing this, will think of a better way to fix this! :-)

PR #1518, opened 2018-03-17
pluginhelper: Install from zip directly

ashthespy:
Optionally install directly from zip, instead of unziping and reziping again.


volumio:
Sorry not merging this: this will ensure we are disciplined in committing all plugins sources, so we get consistent results

ashthespy:
Alright! I will look at trying to get travis to package our zips for us - that should make it less of a chore to keep the zips up to date with the repo source!

volumio:
Amazing! 
Looking forward to automate as much as possible, and make the plugin system more like an app store thing so we can scale, grow and deliver great experieinces... So, feel free to think big!

Issue #1270, opened 2017-07-21
No radio stream title after load playlist


mezcalf:
Saving radio streams as a playlist and reloading produce deleting title of radio streams.

Queue with radio streams:
![before](https://user-images.githubusercontent.com/6116648/28457106-25483f10-6e05-11e7-8be4-999b31304f32.png)

Adding playlist to the queue:
![addtoqueue](https://user-images.githubusercontent.com/6116648/28457124-39947858-6e05-11e7-9d6f-80f1a4e3537b.png)

Queue with radio streams with undefined titles:
![after](https://user-images.githubusercontent.com/6116648/28457114-2ce35840-6e05-11e7-9ffb-4657a6246fc8.png)

Details on forum: https://volumio.org/forum/does-not-show-title-own-web-radio-station-t7109.html

PR #1526, opened 2018-04-02
retain title for webradio

diehardsk:
Issue #1270
Title and albumart of webradio was lost when loading/enqueuing playlist.
Most of detected changes is due to formating consolidation (tab->space) in commonPlayPlaylist function.
Real improvement is on line 176 and 638~668.

### DIFF
--- a/app/playlistManager.js
+++ b/app/playlistManager.js
@@ -173,7 +173,7 @@ PlaylistManager.prototype.enqueue = function (name) {
                         var item = {
                             service: data[i].service,
                             uri: data[i].uri,
-                            name: data[i].title,
+                            title: data[i].title,
                             artist: data[i].artist,
                             album: data[i].album,
                             albumart: data[i].albumart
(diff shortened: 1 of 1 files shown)


xipmix:
Looks ok, I shall test.
I wondered about changing the name of the ```uris``` array to ```array``` but I guess that's not a reserved word or the automatic checks would have complained.
It's nicer for review if the whitespace changes can be a separate commit, but hey ho.

xipmix:
Simple test with one playlist with one item - seems ok on v2.389/rpi2
Tested-by: xipmix@users.noreply.github.com

diehardsk:
Yes I changed that variable name to _array_ so that it is more logical now when having more different fields not only uri. In _PlaylistManager.prototype.enqueue_ it was already used same way.
And about the whitespaces, I'll remeber that for future. I didn't realize it would mess up difference review.
Thank you for testing.

PR #1529, opened 2018-04-07
Don't create a websocket inconditionally at startup

piffio:
Instead of creating the websocket connection at startup for each
invokation, do it only within the 2 methods that currently require
one: `install` and `update`.

This has the nice side effect of not preventing the script to exit
properly when calling any of the other commands, like `refresh`.

Additionally, the use of `process.exit()` is discouraged, and it's better
to just set an `exitCode` so that any unprocessed event in the event loop
will be handled before exiting.

See here https://nodejs.org/api/process.html#process_process_exit_code
for more details.


volumio:
Thanks, well done

Issue #1361, opened 2017-10-08
Player name sometimes changes from Capitalized to lowercase - but not always


2oh1:
In Volumio System settings, Player Name seems to be automatically capitalized.  If you type an all lower case name and hit save, then click to another page (Appearance, for example), when you return to System settings page, the Player Name will have been changed to be Capitalized.

But Airplay sometimes sees the player being broadcast as Capitalized, and sometimes it sees it as lowercase.  This is a problem for two reasons: scripting on the receiving end will fail if the player's name is changed, and also, sometimes (after a save or a restart for example) iTunes will see the changed player name as if it's a new player, while it still waits for the old player.  In other words, iTunes sees the player twice, as if Capitalized and lowercase are two different players, though only one is functional.

While searching, I found issue #357 and #293 but neither has a solution and both are closed.

I'd prefer to have what I enter in System settings as the player name be the player name (Capitalized first latter and lowercase for the rest).

xipmix:
When you are making use of the player name - do you connect it in other ways than with iTunes:  samba (windows shares) ? upnp? ssh connection to the host name? The reason I ask is these methods have various constraints on the name string, some of which may be incompatible with the Capitalized option.

2oh1:
I'm using Volumio strictly as an Airplay receiver.  No windows shares, no upnp.

The only other use I have for the Raspberry Pi Volumio is running on is TriggerHappy.  I'm using TriggerHappy for hotkeys.

I am using ssh, but the only setup I did for ssh was to drop an empty file named ssh into the main boot directory, and I set up ssh-keygen  so that I can send commands to my Mac using TriggerHappy hotkeys.

Once Volumio is up and running, it seems very stable, and the player name is broadcast to Airplay consistently...  but if I have to restart Volumio for any reason, the player name changes to all lowercase even though Volumio System Settings shows it as being capitalized.

Issue #974, opened 2017-01-22
Hostname change incomplete (SMB) & errors


macmpi:
When changing hostname in System UI, SMB service does not get changed to new host name.
Indeed [setHostname](https://github.com/volumio/Volumio2/blob/master/app/plugins/system_controller/system/index.js#L220) does not seem to handle `/etc/samba/smb.conf`, nor restart that service, unlike avahi.

I enclose log of such name change as an example (change from "toto" to "test").
One can note error logs related to `/etc/hosts` permissions, and missing `/etc/avahi/services/volumio.service` (`/etc/avahi/services` directory is indeed empty and `volumio.service` seems to be at `/lib/systemd/system/volumio.service`)

```
Jan 22 19:10:32 toto volumio[789]: info: CALLMETHOD: system_controller system saveGeneralSettings [object Object]
Jan 22 19:10:32 toto volumio[789]: info: CoreCommandRouter::executeOnPlugin: system , saveGeneralSettings
Jan 22 19:10:32 toto volumio[789]: info: System name has changed, restarting Shairport Sync
Jan 22 19:10:32 toto volumio[789]: Discovery: Restarting stuff
Jan 22 19:10:32 toto volumio[789]: Discovery: Stopped ads, if present
Jan 22 19:10:32 toto sudo[1144]: volumio : TTY=unknown ; PWD=/ ; USER=root ; COMMAND=/usr/bin/killall upmpdcli
Jan 22 19:10:32 toto sudo[1144]: pam_unix(sudo:session): session opened for user root by (uid=0)
Jan 22 19:10:32 toto sudo[1144]: pam_unix(sudo:session): session closed for user root
Jan 22 19:10:32 toto sudo[1149]: volumio : TTY=unknown ; PWD=/ ; USER=root ; COMMAND=/bin/chmod 777 /etc/hosts
Jan 22 19:10:33 toto sudo[1149]: pam_unix(sudo:session): session opened for user root by (uid=0)
Jan 22 19:10:33 toto sudo[1152]: volumio : TTY=unknown ; PWD=/ ; USER=root ; COMMAND=/bin/systemctl restart airplay
Jan 22 19:10:33 toto volumio[1063]: src/ohmetacache.cxx:123::dmcacheSaveWorker: can't get task from queue
Jan 22 19:10:33 toto mpd[963]: client: [2] closed
Jan 22 19:10:33 toto sudo[1149]: pam_unix(sudo:session): session closed for user root
[...]

PR #1530, opened 2018-04-09
Fix edge case of playerName with uppercase letters

ashthespy:
Fixes the renaming of device back to default (for the edge case that I had of a playerName in camelCase ).

It would seem that my issues were nothing related to `mdns` and node v8 ;-)

### DIFF
--- a/app/plugins/system_controller/volumiodiscovery/index.js
+++ b/app/plugins/system_controller/volumiodiscovery/index.js
@@ -113,7 +113,7 @@ ControllerVolumioDiscovery.prototype.startAdvertisement=function()
 	console.log("Discovery: StartAdv! " + forceRename);
 	try {
 		var systemController = self.commandRouter.pluginManager.getPlugin('system_controller', 'system');
-		var name = systemController.getConf('playerName');
+		var name = systemController.getConf('playerName').toLowerCase();
 		var uuid = systemController.getConf('uuid');
 		var serviceName = config.get('service');
 		var servicePort = config.get('port');


xipmix:
I'm confused about how this change helps. Can you state what your problem was before and how the change fixes it?
I'm guessing it's because mdns doesn't like camelCase and forcing lowercase avoids the issue, but I could be wrong.

ashthespy:
The [`ControllerSystem.prototype.saveGeneralSettings`](https://github.com/volumio/Volumio2/blob/9ac13199e84bd2c5cf8082565cccc6b80b02df97/app/plugins/system_controller/system/index.js#L225) lets you set an arbitrary playerName, which in my instance was in camelCase. 

However [`volumiodiscovery`](https://github.com/ashthespy/Volumio2/blob/bd5f0420ee53db384140adb5ef1661209efbdea4/app/plugins/system_controller/volumiodiscovery/index.js#L128) the `if ((theName != name) && (!forceRename))` test will fail for this case, resetting the `playerName`

xipmix:
That helps a lot, thank you.

messismore:
> I'm guessing it's because mdns doesn't like camelCase and forcing lowercase avoids the issue, but I could be wrong.

FWIW, "MacBook-Pro", or "iPhone" doesn't seem to be a issue in my network…

It might be a pet peeve, but having Volumio players announce themselves with all lowercase names for Airplay and Spotify Connect irks me more than it should. So inelegant… 🙈

volumio:
playerName is the actual name of the Volumio player

ashthespy:
@volumio Since we derive the hostname anyway from the playerName, can't we allow some flexibility (utf-8 and caps) in it? 
A quick look shows that playerName is utilised mainly by `upnp` , `airplay_emulation`, `networkfs` and `volumiodiscovery` and other plugins (volspotconnect).

volumio:
@ashthespy yes we can, however the hostname should be all lowercase and with - instead of spaces. What's your idea?

ashthespy:
@volumio  How about a validation check in [`ControllerSystem.prototype.saveGeneralSettings`](https://github.com/volumio/Volumio2/blob/9ac13199e84bd2c5cf8082565cccc6b80b02df97/app/plugins/system_controller/system/index.js#L225) with a warning notification if the user tries setting a playerName that doesn't play(hehe) well with Volumio? 
Additionally, we can also push a notification if the parsed hostname is different from the playerName.

xipmix:
+1 for this. Many months ago I tried writing a function to do this and some associated tests but got bogged down with all the name conventions that are allowed. I'll see if I can dig it out and attach here.

ashthespy:
I just quickly looked up the underlying daemons (`upmpdcli` ,` shairport-sync`,`mdsn`) for their respective 'name' parameters, and it looks like there isn't explicit constraints! 
Need to check utf8 support though.

ashthespy:
@volumio a little OT, but is there a specific reason behind displaying [device names in uppercase](https://github.com/volumio/Volumio2/blob/master/app/plugins/system_controller/volumiodiscovery/index.js#L404) in volumiodiscovery?

volumio:
@ashthespy mainly for aesthetics

xipmix:
The 'conventions' for hostname I was referring to were for mdns, netbios, upnp, etc. See #974,  #1361. But one step at at time, let's sort out DNS (and mDNS) first.

ashthespy:
A small summary of my findings so far. I found `playerName`, `system.name`, `system_name` and `hostname` which each seem to be used in different places.

`system.name` used in:
```
app/plugins/music_service/airplay_emulation/index.js:  var name = this.commandRouter.sharedVars.get('system.name');
app/plugins/system_controller/networkfs/index.js:      var systemShare = self.commandRouter.sharedVars.get('system.name').toUpperCase();
app/plugins/system_controller/system/index.js:         this.commandRouter.sharedVars.addConfigValue('system.name', 'string', self.config.get('playerName'));
app/plugins/system_controller/system/index.js:         self.commandRouter.sharedVars.set('system.name', player_name);
app/plugins/user_interface/websocket/index.js:         var name=self.commandRouter.sharedVars.get('system.name');app/plugins/user_interface/websocket/index.js:                          var name = self.commandRouter.sharedVars.get('system.name');

```

`playerName` Used in:
```
app/index.js:   var name = file.playerName.value;
app/plugins/audio_interface/upnp/index.js:               var nameraw = systemController.getConf('playerName');
app/plugins/system_controller/networkfs/index.js:        var nameraw = systemController.getConf('playerName');
app/plugins/system_controller/volumiodiscovery/index.js: var name = systemController.getConf('playerName').toLowerCase();
app/plugins/system_controller/volumiodiscovery/index.js: systemController.setConf('playerName', theName);

```
`system_name` used locally in `wizard`: 
```
app/plugins/miscellanea/wizard/config.json:  "system_name":{"value":"Volumio","type":"string"}
app/plugins/miscellanea/wizard/index.js:     var systemName = self.config.get('system_name', 'Volumio');
``` 

I don't fully understand the difference b/w `playerName` and the `system.name` but can we use the `playerName` for things that are more flexible  (mDNS, display, etc) and use `system.name` as the sanitised name that is used for `netbios`, `upnp`, `hostname`, etc 
I started writing some regex for this, but don't know who all will play nice with utf8. 
Would https://github.com/bestiejs/punycode.js be a more straight forward option for the sanitising then?

xipmix:
I dug out the stuff I wrote a while ago. It's incomplete and probably the wrong direction but maybe the tests I wrote could be reworked into something useful. See attached. Licence: public domain.
[hostnamecheck.txt](https://github.com/volumio/Volumio2/files/1914849/hostnamecheck.txt)
[hostnamecheck-tests.txt](https://github.com/volumio/Volumio2/files/1914851/hostnamecheck-tests.txt)

I don't know what to advise about depending on punycode. I would suggest getting something that works with the ascii subset, with the intention of adding utf8 support in another iteration.

ashthespy:
Sorry, github didn't send me any emails for some reason.

@xipmix That is quite extensive and complete, I was thinking of a more simple Ok/NotOk kind of test for the hostname and defaulting to `volumio`, but your tests make it easier for sure!

I made some preliminary tests with punycode for utf-8 hostnames, and it's quite network depended I feel. My browser could find `volümiö.local` (by automatically converting it to `xn--volmi-nua3b.local` but ping and unpnp had some issues. 

Given the broad user base of Volumio, I would again suggest we use the `PlayerName` externally as the 'pretty name' and use the `system.name` internally (mDNS, upnp, hostname)

If everyone is along the same lines, I shall make a small POC the following days :-)

volumio:
@ashthespy Hi Ash, sorry for this super late reply...
Yes this issue is annyoing, so looking forward for your POC!

messismore:
Has this been fixed?

barereef:
@volumio This is still an open issue in 2.917.

messismore:
Fixed in Volumio 3 🥳

PR #1533, opened 2018-04-10
First Intel Atom Baytrail/ Cherrytrail sound support

gkkpch:
Currently experimental only.
The Intel SST and HDMI/DP LPE Audio devices need to be treated a little different.
bytcr-rt5651 needs a default mixer and HDMI is on hw:2,0, currently not clear what hw:0,0 and hw:1,0 are meant for (Analogue out and Display Port)?


volumio:
Well done Gé!

PR #1555, opened 2018-05-28
Enforce leading / on NFS path supplied by user

xipmix:
NFS mounts require an absolute path to the exported directory on the server.
The volumio UI does not encourage this, as the default is mouting via CIFS.
To avoid troubling the user with these details, fix up the path before the
configuration is used or saved to disk.

Motivation: volumio/Volumio2-UI#511.

### DIFF
--- a/app/plugins/system_controller/networkfs/index.js
+++ b/app/plugins/system_controller/networkfs/index.js
@@ -379,6 +379,15 @@ ControllerNetworkfs.prototype.addShare = function (data) {
 		path = path.replace(/\/+/g,'/');
 		path = path.replace(/^\//,'');
 	}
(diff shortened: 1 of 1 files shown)


app/plugins/system_controller/networkfs/index.js
```diff
@@ -379,6 +379,15 @@ ControllerNetworkfs.prototype.addShare = function (data) {
 		path = path.replace(/\/+/g,'/');
 		path = path.replace(/^\//,'');
 	}
+	if (fstype == 'nfs') {
```
chris-rudmin: It is good practice to prefer strict equals `===`.

volumio:
Thanks to both

PR #1557, opened 2018-06-02
czech translation update

manin22:
I don't like localizations at all. Anyway, it should be updated if it has been once created, so here you are my small contribution.


volumio:
Děkuji!

manin22:
:-) You just made my day

Issue #1548, opened 2018-05-12
undefined language_code causing app to crash


SylannBin:
Original context:

The Volumio webapp stopped working several seconds after I reached the page for the first time on this install.

While connected to the rpi via ssh, I tried:

    volumio status  // 

then I tried

    volumio vrestart //

After spending a lot of time trying to find problems in my rpi, I decided to check volumio itself.

I have then found this recurring error in the log:

```
Error: /volumio/app/i18n/strings_undefined.json: ENOENT: no such file or directory, open '/volumio/app/i18n/strings_undefined.json'
    at Error (native)
    at Object.fs.openSync (fs.js:641:18)
    at Object.fs.readFileSync (fs.js:509:33)
    at Object.readFileSync (/volumio/node_modules/jsonfile/index.js:67:22)
    at CoreCommandRouter.loadI18nStrings (/volumio/app/index.js:1566:25)
    at new CoreCommandRouter (/volumio/app/index.js:65:10)
    at Object.<anonymous> (/volumio/index.js:42:21)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.runMain (module.js:604:10)
    at run (bootstrap_node.js:389:7)
    at startup (bootstrap_node.js:149:9)
    at bootstrap_node.js:504:3
```

Starting the app myself with `/volumio/index.js` gave this exact same error.

Looking at the code I eventually reach the part CoreCommandRouter.loadI18nStrings part...

And here is what I see:

```
    var self=this;
    var language_code=this.sharedVars.get('language_code');

    this.logger.info("Loading i18n strings for locale "+language_code);

    this.i18nStrings=fs.readJsonSync(__dirname+'/i18n/strings_'+language_code+".json");
    this.i18nStringsDefaults=fs.readJsonSync(__dirname+'/i18n/strings_en.json');

    // ...
```
Since `language_code` is not defined in the `sharedVars` (for some reason), the local variable of the same name is also undefined. And this part of the program tries to read a file named `/i18n/strings_undefined.json`

Reading a file synchronously and not handling an error that begs to occur kind of strikes me.

xipmix:
Great analysis. What language did you configure please?

volumio:
[...]

PR #1566, opened 2018-06-22
Avoid crashing if locale strings file is missing

xipmix:
Issue #1548 presented a case where the sharedVars language_code
was an unexpected value and caused the system to crash.
This patch handles the simple cases - language_code undefined
and/or missing strings_<foo>.json file. Other kinds of errors could
still cause a crash.

Tested by setting language to French and renaming the app/i18n/strings_fr.json so file so it can't be found.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -1579,7 +1579,11 @@ CoreCommandRouter.prototype.loadI18nStrings = function () {
 
         language_code = 'en';
         i18nFile = __dirname + '/i18n/strings_' + language_code + '.json';
-        this.i18nStrings = fs.readJsonSync(i18nFile);
+        try {
+            this.i18nStrings = fs.readJsonSync(i18nFile);
+        } catch( e ) {
+            this.logger.error('File ' + i18nFile + ' not accessible, no fallback available');
+        }
     }
 
     var categories=this.pluginManager.getPluginCategories();
--- a/app/index.js
+++ b/app/index.js
@@ -1581,6 +1581,10 @@ CoreCommandRouter.prototype.loadI18nStrings = function () {
         i18nFile = __dirname + '/i18n/strings_' + language_code + '.json';
         try {
             this.i18nStrings = fs.readJsonSync(i18nFile);
+            // Flag the fallback to 'en' in the UI, to reduce confusion.
+            // NB we can't use the getI18nString method yet;
+            //    we must display untranslated strings.
+            this.pushToastMessage('warning', 'Language File Missing', 'Falling back to English' + ' ' + error.message);
         } catch( e ) {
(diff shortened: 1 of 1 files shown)


app/index.js
```diff
@@ -1562,10 +1562,27 @@ CoreCommandRouter.prototype.loadI18nStrings = function () {
...
+
     this.logger.info("Loading i18n strings for locale "+language_code);
 
-    this.i18nStrings=fs.readJsonSync(__dirname+'/i18n/strings_'+language_code+".json");
-    this.i18nStringsDefaults=fs.readJsonSync(__dirname+'/i18n/strings_en.json');
+    try {
+        this.i18nStrings = fs.readJsonSync(i18nFile);
+    } catch( error ) {
+        if (error.code === 'ENOENT') {
+            this.logger.info('Warning: file '+i18nFile+' not found, defaulting to '+i18nDefaultFile);
```
volumio: This will cause a crash, since i18nDefaultFile is not defined, if you remove this reference it will work

xipmix: Thank you I spotted that later myself. I did tes,t but somehow did not test this final version. I tried just dropping the reference but then when reading the log I wanted to know what volumio had done about the problem it reported. So I opted to define i18nDefaultFile.

app/index.js
```diff
@@ -1562,10 +1562,28 @@ CoreCommandRouter.prototype.loadI18nStrings = function () {
...
+        this.i18nStrings = fs.readJsonSync(i18nFile);
+    } catch( error ) {
+        if (error.code === 'ENOENT') {
+            var i18nDefaultFile = __dirname + '/i18n/strings_' + 'en' + '.json';
+            this.logger.info('Warning: file '+i18nFile+' not found, defaulting to '+i18nDefaultFile);
+            language_code = 'en';
+            i18nFile = __dirname + '/i18n/strings_' + language_code + '.json';
+            this.i18nStrings = fs.readJsonSync(i18nFile);
+        } else {
+            throw error;
```
volumio: While we're at it: why don't we catch the error not only for the ENOENT error? This will make it more robust. What do you think?

xipmix:
Ok I reworked this further and may possibly have gone too far... 
You could drop the commits after a71fc0e with no ill effects. I tested that and then the end product, including removing the strings_en.json file.

xipmix:
I think 430cadf857c012fabf56010324ffdc499838cede obviates this now.
What do you think about displaying a popup if we have to fall back (3edcf8812d95241cf84dd5d8e955c6640a5f9d4e) ?

volumio:
@xipmix I see no reason to display a popup, since this might raise a warning on the user but he has no way of fixing the situation

PR #1580, opened 2018-07-06
Extend play command in websocket API.

morphZ:
Hi,

I hit the problem to select a radio stream from the playlist in volumio. Therefore I had to add the possibilty to address a specific item via websocket. As I couldn't see which Object you usally use, that has ` .value` property, I converted the recieved message into a primitive number.

This works for me. See if it works for the main project as well, as I don't in what other ways the API is used.

I appreciate any comment on this :)

morphZ


volumio:
Hi!
Unfortunately this will break many things, since we always send object.
What you want is send:
'play' , {"value":2}

PR #1583, opened 2018-07-14
Fix timer overrun bug & prefetchDone not reset bug

lostmyshape:
Fixes prefetching too many times due timer overrunning when buffering track & makes sure prefetchDone variable is always reset when the timer is reset to prevent unwanted behavior when queue has 2 consecutive tracks from different music services.

Full explanation in forum: https://volumio.org/forum/bugs-statemachine-increaseplaybacktimer-method-t10340.html


volumio:
Thanks!

PR #1596, opened 2018-08-24
Suggested patch for Debian Stretch

ggiraudon:
Fixed network status display. Ifconfig output format has changed.


volumio:
Thanks, I took inspiration from your PR to get something "universal" that will work both on jessie and stretch. See https://github.com/volumio/Volumio2/commit/0419f8294c6ab087aab092734b6cd8dd82980b77

Issue #1612, opened 2018-09-19
Consistant code style


ashthespy:
As Volumio grows, so does it's codebase. 
Going by this nice quote from [`idiomatic.js`](https://github.com/rwaldron/idiomatic.js/):
>All code in any code-base should look like a single person typed it, no matter how many people contributed.

Would there be a Volumio specific style guide? If not, [`standardjs`](https://github.com/standard/standard) seems to be the most widely used, and has wide IDE support!

volumio:
Good idea. Would you like to help setting it up and fixing code?

volumio:
I feel silly, but... https://github.com/Flet/semistandard ?

Hate to say goodbye to semicolons...

ashthespy:
Haha, I just linted some code and it felt strange not to have semicolons. 

In terms of workflow, we could add a CI to run and commit the lints, or use some git hooks to run `standard --fix` before any commits. But then in depends on the devs individual workflow.

volumio:
Deal! Semistandard and githooks!

xipmix:
+1```;```

PR #1613, opened 2018-09-20
Set up consistant codestyle

ashthespy:
Further to #1612 

### Workflow 
First install the dependencies with `npm install --only=dev` or more specifically:
   ```shell
   > npm i husky@next lint-staged prettier semistandard
   ```
All further commits will be automatically formatted using the `semistandard` style  and will prompt you to adhere to the styleguide. 
We could then fix all lint errors in the file when you make some changes, rather than attempt to fix the entire repo in one go.  
Of course you can always pass the `--no-verify` flag to git if you don't feel inclined to fix all the lint errors, and just use the formatting.  

Also included sample of the formatting for the parent dir. I started fixing all the lint errors, but soon got bored. :-}

If this works, can run the formatting for the entire repo in oneshot.

### DIFF
--- a/crashreport.js
+++ b/crashreport.js
@@ -2,59 +2,58 @@ var moment = require('moment');
 var execSync = require('child_process').execSync;
 var exec = require('child_process').exec;
 var argv = require('yargs').argv;
-var logFile = "/tmp/crashdump";
+var logFile = '/tmp/crashdump';
 var lastCrashDescription = '/tmp/lastcrash';
 var description = 'Unknown';
 var fs = require('fs-extra');
 
-
 if (process.argv[2] != undefined) {
-    description = "'" + process.argv[2] + "'";
+  description = "'" + process.argv[2] + "'";
 }
 var now = moment().format('YYYY-MM-DD HH:mm');
 var sinceTime = moment().subtract(1, 'minutes').format('YYYY-MM-DD HH:mm');
-var minuteDump = execSync('/usr/bin/sudo /bin/journalctl --since="' + sinceTime +'" > /tmp/crashdump', {uid: 1000, gid: 1000});
+var minuteDump = execSync('/usr/bin/sudo /bin/journalctl --since="' + sinceTime + '" > /tmp/crashdump', {uid: 1000, gid: 1000});
 var releaseDump = execSync('cat /etc/os-release >> /tmp/crashdump', {uid: 1000, gid: 1000});
 var variant = getVariant();
 
 try {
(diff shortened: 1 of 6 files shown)


volumio:
Great, one thing: can we just remove prettier? I find it rather annoying, especially when it breaks lines. What do you think?

ashthespy:
I guess so, the main thing that prettier brings to the table is wrapping of long lines, but it looks to be quite aggressive at it.

volumio:
Ok, so can you remove prettier and redo the PR? The code has been reformatted with prettier so I guess you can just stash it and I will take care of initial formatting

ashthespy:
Will rework and force push in a bit..

ashthespy:
Done - also ran `--fix  *.js` for  `__dirname`. 
Did not fix the errors the linter warns about, I'd propose we fix the formatting in one commit, and the linting in another?

volumio:
Great! It looks really really nice!

ashthespy:
In the near future (once all errors are also fixedup), we can set up travis to complain if any PR's/commits don't match this style..

sstark:
Is there a document somewhere that explains the official volumio coding style?

ashthespy:
From #1612 [`semistandard`](https://github.com/Flet/semistandard) style was settled upon. It is [`standardjs`](https://standardjs.com/) but with semicolons. 

PS: Relevant kxcd https://xkcd.com/927/

!xkcd

sstark:
Thanks, found it on the standardjs link. I wanted something I can read and learn and then lint later. There is no mentioning of tabs vs. spaces. I don't care which, but maybe there is a standard for the project?

ashthespy:
Hmm, https://standardjs.com/rules.html - first rule ;-)
What editor/IDE do you use? You should find an appropriate package/plugin for it at [`semistandard`](https://github.com/Flet/semistandard#editor-plugins) readme page.

sstark:
Hm, I just clicked that first links, searched for "tabs" and found: "just pick something" :) Anyway, that's good. I use vim which has all the plugins needed, but am completely new to js and when making a patch for the ampswitch plugin I found completely messed up code that I want to fix.

ashthespy:
FWIW - the idea behind picking this is that you don't have to focus on setting this all up manually, and just use the inbuilt npm package `semistandard`. But I do understand that if you don't have a `nodejs` installation then it a bit much to install `nodejs` just to lint ;-)

FWIW (II) - The plugin repo doesn't currently have any style guide - I think when I last edited `ampswitch` I just used atom's Js formatter

PR #1633, opened 2018-10-21
Search and play non-VA albums that contain multiple artists

hiisukun:
Often when index.js checks to see if we are playing a compilation, but we are not, it then relies on the 'artist' field of tracks to decide if they are part of the same album to display or play. This patch makes these occasions rely instead on 'albumartist', which should hopefully correctly play more albums.

For example, if an album by 'Sting' contains a track with the artist listed as 'Sting feat. Beethoven', when you find that album under the artist view, it would previously not list the track featuring Beethoven. After this patch, it would list all the tracks on the album.

This should (at least partially) resolve:
https://github.com/volumio/Volumio2-UI/issues/448

### DIFF
--- a/app/plugins/music_service/mpd/index.js
+++ b/app/plugins/music_service/mpd/index.js
@@ -2061,12 +2061,7 @@ ControllerMpd.prototype.explodeUri = function(uri) {
         var safeArtistName = artistName.replace(/"/g,'\\"');
         var safeAlbumName = albumName.replace(/"/g,'\\"');
 
-		if (compilation.indexOf(artistName)>-1) {  //artist is in Various Artists array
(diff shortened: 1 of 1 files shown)


hiisukun:
My volumio version is a little older than this repo - so I've ported in just the changes I needed to make index.js work with my 'mixed artist but non-VA albums'. Unfortunately this means I can't simply throw this whole file into my setup to test it - sorry!

hiisukun:
Come to think of it, perhaps this could probably do with some fallback support: If albumartist property is null, use artist instead.

volumio:
Yes, please add the fallback support, after I can review it and hopefully merge

PR #1652, opened 2018-11-16
Improve quoting of user input to mountShare

xipmix:
Since the mounting of NAS drives is done as root, we need to take
some care with quoting the strings supplied by the user for things
like username, password, mount options, etc. As long as the strings
are single-quoted, the shell used for calling 'mount' should never
try to interpret the characters in between. Things get trickier when
the input shell contains a single quote. Then we have to turn any
single ' within the string into the sequence '"'"'.


volumio:
Well done, thanks

PR #1656, opened 2018-11-20
Added ability to control the visibility of built-in browse sources (e…

srhamil:
….g. Genres)

so that users that don't use a local music libary, don't use all the ways of browsing, and/or have small displays
can control the choices available on the Browse pane.
image

This is managed b a control panel extension:
image

### DIFF
--- a/app/i18n/strings_en.json
+++ b/app/i18n/strings_en.json
@@ -225,7 +225,8 @@
     "ALBUMART_RESET_CACHE":"Reset Album Art cache",
     "ALBUMART_RESET_CACHE_DOC":"The system caches your albumart to improve performance, but this prevents newly added albumart from being shown. By resetting the cache you will be able to see newly added albumart",
     "METADATA_IMAGE":"Show embedded albumart",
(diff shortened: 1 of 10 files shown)


Mad-M4x:
Does it Works under V2.526 ?
If not, can you fixed it please?

srhamil:
I asked the core development team to integrate this change into volumio. So far as I know this has not been done and this feature only works in my development build. sorry!
r/
srhamil
    On Friday, January 25, 2019, 7:22:40 AM HST, Mad-M4x <notifications@github.com> wrote:  
 
 
Does it Works under V2.526 ?
If not, can you fixed it please?

volumio:
The code you submitted requires a bit of refactoring. It's on our TODO list, unfortunately we did not have much time to look into it.

What we plan to do:
- When you need to set the visibility, just fetch the browsesources object from musiclibrary.
- If a file called, browseSourcesVisibility is found, match the value of each object with the ones from musiclibrary and send it back
- When saving, simply update the browseSourcesVisibility file.
- When calling getbrowsesources, do the same

volumio:
I've implemented it via https://github.com/volumio/Volumio2/commit/483b493a2b6e2f9a9c331b59b01094a58ba670dc

PR #1677, opened 2018-12-13
Small spelling and grammar issues

xipmix:
balbuze's recent work prompted me to go and look at strings_en.json.

Not actually tested on a live system but I didn't touch any quote marks so should be ok.


volumio:
Thank you very much, super appreciated

PR #1694, opened 2019-01-02
There is no readJSONSync() function

xipmix:
Not sure if this matters but it's inconsistent which just wastes mental bandwidth. Every other call is readJsonSync().

Tested(?) on 2.513 by switching language to DE.


volumio:
Thanks, well spotted!

Issue #1701, opened 2019-01-10
API for currently playing playlist


mrsimpson:
## Motivation

I'm implementing an NFC-reader plugin which shall make it possible to start a particular playlist. 

## Issue 

Since I want to make is as easy as possible for the user to bind a playlist to a recognized tag, I'd like to retrieve  the name of the currently playing playlist. But there seems to be no API for that

## Suggestions

I can imagine multiple solutions:

- Add the name of the playlist to the `status`
- Provide a dedicated read-API on Websocket `currentPlaylist`

Workaround: Add a mechanism for a plugin to hook into the `playPlaylist` event - this way, my plugin could keep track of the playlist playing itself. Is there such an option?

PR #1702, opened 2019-01-11
Propagate playlist playing via Websocket

mrsimpson:
Provides a workaround #1701

This enhancement allows plugins and other consumers to visualize the currently playing playlist.
Note that consumers need to react on other state changes as well if they need to know once the playlist is "switched" by playing another title directly.


mrsimpson:
I know I should create a PR to `develop`. I will update the PR once I got your feedback. 
Until then, having it based upon `master` makes it easier for myself

mrsimpson:
Sorry for the bad first-contribution-impression! I was not aware of my Editor having formatted the code on save.

> There is work going on to standardise whitespace (#1613) - perhaps worth checking if your changes are following that.

@xipmix Thanks! I did this, but actually there were even more semistandard issues  in this file and others. I ran `npm run lint` and got some 71 files changes. Thus, I reverted the previous commit, switched off "lint-on-save" and got the smaller differ

> Also I find the second sentence of your comment (Note that consumers...) difficult to understand. 
You seem to be saying that if consuming the playlist via the websocket, clients need to keep an eye on other parts of the volumio state or they will get confused. Can you clarify this?

Hmm - let me try to rephrase: This PR simply notifies a "playPlaylist" via Websocket. But this is not equal to "a song of a playlist is playing. 

- an invalid playlist name may be requested
- After the playlist has been played, another music service may be requested to play.

Thus, if a consumer wants to know whether what's playing is part of a playlist, it also needs to watch `play`.

Was this understandable?

xipmix:
That's easier to see, thanks. It's up to @volumio to decide about merging it. Also your explanation helped me understand more.

I struggle with how the control flows here so my question may be wrongheaded -
have you thought about checking the return value of ```self.commonPlayPlaylist(self.playlistFolder, name)``` to see if the playlist was a valid name?

The other question I had was how monitoring ```play``` helps - how do you tell whether what is currently being played is part of a playlist or not?

mrsimpson:
P.s.: In other words: I was about to just create an issue saying "playlist playing not exposed in status", but then I found the option to make volumio pull another repo on the command line - so I thought to find a solution myself. Once I had done that, I thought it might be useful to others.

PR #1709, opened 2019-01-22
fix: exclude hidden folders from directory browsing

davidcole1977:
## What

- When a `browseLibrary` request is received, exclude folders beginning with `.` from the array that's returned to the client.

## Why

- Folders hidden with `.` are hidden system folders and are normally not available to the user. These include directories like `.Trashes` that can easily be included on the user's media drive without them being aware of it. It's very strange to see these displayed to the user. Here's an example:

volumio-library-hidden-folders

## Solution

- The solution is crude and inelegant, but the surrounding code in `mpd/index.js` is not the easiest to work with, so I went with something very simple that won't require any refactoring to implement.

### DIFF
--- a/app/plugins/music_service/mpd/index.js
+++ b/app/plugins/music_service/mpd/index.js
@@ -1271,13 +1271,17 @@ ControllerMpd.prototype.lsInfo = function (uri) {
                             }
 
                             name = namearr.pop();
-                            list.push({
-                                type: dirtype,
-                                title: name,
-                                service:'mpd',
-                                albumart: albumart,
-                                uri: s0 + path
-                            });
+
+                            // exclude hidden folders (eg. ".Trashes")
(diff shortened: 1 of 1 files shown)


xipmix:
This looks fine to me but I wonder if it is simpler still to check for the leading '.' and if found, skip to the next loop iteration ('continue').

davidcole1977:
@xipmix Bingo! You know I'm so used to using the newer array methods to iterate over arrays I had forgotten about `continue`. I feel slightly embarrassed ;)

volumio:
Do you want to push this impro? IMHO we could merge as it is ;)

davidcole1977:
@volumio @xipmix that amendment has been made in 0151164 - the reason I changed it from using `.pop()` to get the `name` is because `pop` mutates the original array, which is used in the subsequent code.

PR #1717, opened 2019-02-14
Implemented Concert Mode in the backend

eisen:
**Added a Concert Mode to Volumio**

- New variable concertMode to the State Manager.
- New endpoint concert-mode to the RESTful API
- New event concert-mode to the Socket.io interface
- In play mode and if Concert mode is enabled, when a new song is loaded or the user requests the next song, play #mode will be set to pause

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -1425,6 +1425,35 @@ CoreCommandRouter.prototype.repeatToggle = function () {
     return self.stateMachine.setRepeat(repeat, false);
 }
 
+CoreCommandRouter.prototype.volumioConcertMode = function (concertMode) {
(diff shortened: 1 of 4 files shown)


volumio:
Very good pull request, congratulations! However I don't understand the purpose of concert mode. Would you mind explaining me a bit more?

eisen:
This is a very niche option. My wife is a singer and we use Volumio as our track player during concerts. Right now I need to stop a song right before the next one starts. With this mode enabled, once a song finishes playback, the next song is cue’d and playback is paused.

volumio:
Sorry, but after some internal discussions we cannot merge, as for a very nieche function we need to offer a big overhaul of the UI.

PR #1721, opened 2019-02-22
Fix track title display in music browser when tracknumber metadata doesn't exist

futuralogic:
**Scenario**: 
In Music Library Settings: _Show Track Numbers_ is **enabled** and a file doesn't have tracknumber metadata. The resulting entry in the music browser displays as a single hyphen "-". The filename does come through correctly when _Show Track Numbers_ is off.

The crux of the code is this snippet, which is laced through several of the MPD music service methods:

```
if (!tracknumbers) {
    var title = self.searchFor(lines, i + 1, 'Title:');
}
else {
    var title1 = self.searchFor(lines, i + 1, 'Title:');
    var track = self.searchFor(lines, i + 1, 'Track:');
    var title = track + " - " + title1;
}
... // a few lines later...
// Subsequently the title will be either the composite title from above, or 'name' which is the filename.
if (title) {
    title = title;
} else {
    title = name;
}
```

The change I suggest is to revise to:
```
if (!tracknumbers) {
    var title = self.searchFor(lines, i + 1, 'Title:');
}
else {
    var title1 = self.searchFor(lines, i + 1, 'Title:');
    var track = self.searchFor(lines, i + 1, 'Track:');
    if( track && title1 ) {
        var title = track + " - " + title1;
    } else {
        var title = title1;
    }
}
```

This results in the title being set correctly, to the filename, when the tracknumber or title meta data does not exist.

### DIFF
--- a/app/plugins/music_service/mpd/index.js
+++ b/app/plugins/music_service/mpd/index.js
@@ -1278,7 +1278,7 @@ ControllerMpd.prototype.lsInfo = function (uri) {
                             } else {
                                 dirtype = 'folder';
                             }
-                            
+
                             list.push({
                                 type: dirtype,
                                 title: name,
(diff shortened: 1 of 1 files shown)


volumio:
Thanks!

Issue #1724, opened 2019-02-23
Errors from UPNP browser service in systemd log


SteveTalbot:
I'm seeing this error lots of times in the systemd log (journalctl -f):

```
error:  message=Cannot read property '0' of undefined, stack=TypeError: Cannot read property '0' of undefined
at /volumio/app/plugins/music_service/upnp_browser/index.js:83:67
at /volumio/app/plugins/music_service/upnp_browser/index.js:626:21
at Parser.<anonymous> (/volumio/node_modules/xml2js/lib/parser.js:303:18)
at emitOne (events.js:116:13)
at Parser.emit (events.js:211:7)
at SAXParser.onclosetag (/volumio/node_modules/xml2js/lib/parser.js:261:26)
at emit (/volumio/node_modules/sax/lib/sax.js:624:35)
at emitNode (/volumio/node_modules/sax/lib/sax.js:629:5)
at closeTag (/volumio/node_modules/sax/lib/sax.js:889:7)
at SAXParser.write (/volumio/node_modules/sax/lib/sax.js:1436:13)
at Parser.exports.Parser.Parser.parseString (/volumio/node_modules/xml2js/lib/parser.js:322:31)
at Parser.parseString (/volumio/node_modules/xml2js/lib/parser.js:5:59)
at /volumio/app/plugins/music_service/upnp_browser/index.js:625:24
at Request.handleRequestResponse [as _callback] (/volumio/node_modules/unirest/index.js:463:26)
at Request.self.callback (/volumio/node_modules/unirest/node_modules/request/request.js:187:22)
at emitTwo (events.js:126:13)
```

I was able to reliably trigger the error by browsing to "Media Servers" in the GUI, but it seems to appear randomly at other times too.  In my case, the problem is caused by Volumio discovering a Philips Hue bridge on the local network, and the Hue bridge doesn't have a serviceList.

It looks like a relatively simple fix to harden the code to cope with devices like this.  Pull request to follow.

volumio:
Very well done

PR #1725, opened 2019-02-23
Deal with devices without a serviceList #1724

SteveTalbot:
This is a suggested fix for #1724


volumio:
Thanks Steve, like your coding style very much ;)

PR #1727, opened 2019-02-26
Fix `undefined` mixer when mixer type is None

ashthespy:
Currently `alsa.outputdevicemixer` returns an `undefined` when Mixer type is set to `None`.


volumio:
Thanks!

Issue #1716, opened 2019-02-13
Adding album to playlist adds entire Artist discography


livieris:
V 2.522.  Raspberry pi 3B
I also tried it with V 2.527.
When I try adding an album of an artist to new or existing playlist, it'll add the entire artists discography. The only work around is to add each individual song.
To reproduce, select an artist with more then one album, bring up the quick menu for an album, add to playlist. It doesn't matter if the playlist exists or if it's a new one. After adding, go to that playlist, you should see every album for that artist.

xipmix:
Thanks for reporting this, I will try to reproduce. Are you aware of a version where this did not occur?

xipmix:
Tested on 2.513 and it doesn't do that.
 * I had a small library on a NAS with two Van Morrison albums (and some other artists).
   * Selected one album, add to a new playlist. Works as expected, just adds the tracks from that album.
   * Deleted playlist, selected other album, added to a new and differently named playlist. Works as expected, just adds the tracks from that album.
   * Tried one album with an existing playlist. Works as expected, just adds the tracks from that album.

Can you please add [Regression] to the title of this issue?

volumio:
Might be due to this commit: 
https://github.com/volumio/Volumio2/commit/73603df2a9106c7b1eac6b3cc86c9b00610d37a9

@xipmix could you please check by reverting such commit?

xipmix:
I have 2.513 running and tried _applying_ that commit. I don't see the behaviour. I have to stop now.

When I get a little time I will try reverting it in 2.527, @livieris please feel free to try yourself and letting us know.

livieris:
I'll add some more info that I left out because I didn't think it would matter to this case.  My entire library is on a flash drive, ~17,000 songs. I am also using it with a waveshare 7in touchscreen.

xipmix:
Tried to reproduce again, with a fresh 2.526 image. Can't do it.
 * Set up my test music collection, let the scan complete.
 * Added one album to a new playlist that did not exist before I started. Playlist shows just the content of that album. 
 * Added a different album (same artist) to a different playlist. Playlist shows just the content of that album.
 * cleared all but one song from one of the new playlists above and tried adding an album again (same artist) to that existing playlist. Playlist got just the content of that album added.

Hm. The large collection is the most significant difference between our setups. Can you make a smaller drive with say 3 artists and a few albums each? That would rule the large collection out and make analysing logs more tractable. If you can reproduce with a smaller collection, can you send some logs?
[...]

PR #1728, opened 2019-02-27
Correct handling of artists:// uri with an album appended

xipmix:
Fixes: #1716

Tested on rpi2 and release 2.526.


xipmix:
PS I have not tested for any regressions in handling other use cases of artists: uris.

volumio:
Thanks

PR #1741, opened 2019-03-24
Update dacs.json

valkyee121:
90x8


volumio:
But the drivers are not included in the distro, so it is useless

valkyee121:
ok.Em..Is there any plan to integrate these drivers for sparky？

volumio:
No

Issue #1747, opened 2019-04-07
Mixer Type undefined Issue


thegunslingers:
Setup: 
Volumio on a PiZero
USB DAC 
Hardware mixer type

Everything works as it should when first configured, but occasionally with the USB DAC turned off the mixer type will be changed to undefined and requires an edit to the playback options to fix. I have tested restarting Volumio (with the DAC off) and verified that is one case where the mixer type changes from hardware to undefined and Mixer Control Name changes to a null value.

When Mixer Type is changed to undefined, other settings like Output Device and general playback options are not changed. Audio Playback continues to work while mixer type is set to undefined (volume control doesn't work clearly though).

xipmix:
So ... you have a hardware mixer, you configure volumio to use it and then you(?) turn it off. If not by you, how does it get turned off - a powersave feature?

If you turn the DAC on again and bring volumio up, what happens? 
It sounds like volumio does not notice, and does not fix the Mixer Type.

What do you think should happen here - should volumio automatically change to the software mixer? Or should Mixer Type change to some other defined value, e.g. 'not_detected' ?

thegunslingers:
apologize for the delayed response.
If I get Mixer Type to undefined, then turn on the dac then turn on volumio, mixer state stays as undefined, like you were thinking.
Ideally in my opinion it would stay 'hardware mixer' with 'mixer control name' saved as it was set until manually changed to be something else, even if the mixer is not detected or not plugged in.  I think it would be a little more robust with this setup.  And bad boot up order, or turning things off would not force a settings adjustment.

Thank you!

thegunslingers:
I have been working on a fix for this issue, and I have some working code, but I do have one JS question, before I try to submit the fix.
I want to check the config file to ensure mixer is not undefined or empty, I looked at other sections of the code and followed what I saw. The if statement appears to execute everytime though even when my alsa config.json has
`  "mixer": {
    "type": "string",
    "value": ""
  },`
`ControllerAlsa.prototype.setDefaultMixer  = function (device) {
	var self = this;
        var defaultmixer = self.config.get('mixer', '')
        if(defaultmixer.value != undefined && defaultmixer!=''){ 
                    ...js fix here...
        }`

thegunslingers:
Submitted pull request with fix

xipmix:
Hi, is this resolved for you - if so could you close it please?

thegunslingers:
thanks the hot plug changes now fix this issue

PR #1759, opened 2019-05-18
Mixer Type Undefined Issue

thegunslingers:
In regards to my issue: #1747 

I have added a new switch to the ALSA_Controller UI. It allows users with hardware mixers to not have to worry about the boot order between Voluimo and a USB DAC or other hardware mixer. In the current code if Volumio is turned on before the DAC the Mixer settings are cleared and have to be reconfigured manually.

- The siwtch is displayed when Mixer Type is set to Hardware
- When enabled saved the Mixer Type and Mixer (Don't allow them to be set to undefined on reboot if the external mixer is not turned on)


volumio:
Thanks, but a recent commit solves this issue (USB DAC hotplug)

PR #1760, opened 2019-05-24
Add filter "Albums by Composer"

Olchie:


volumio:
Hi @Olchie . I cleaned up the code, it's a good start.
However the composers cannot be played (I commented the albumartist part because otherwise it would have made no sense to have a special composer facility).
Please review it.
See my edits:
https://github.com/volumio/Volumio2/pull/1812

PR #1763, opened 2019-06-01
Feature/restapi

fanciulli:
Introducing RESTAPI for browsing and playback


app/plugins/user_interface/rest_api/browse.js
```diff
@@ -0,0 +1,361 @@
...
+    if(path=='/')
+    {
+        var content={"navigation":{
+            "lists":[{
+                "items":[]
+            }]
+        }};
+
+
+        content.navigation.lists[0].items.push({"title":"MusicLibrary",
```
volumio: @fanciulli We should not hardcode those, but take them from the getBrowsesources:
https://github.com/volumio/Volumio2/blob/master/app/index.js#L236

This way you'll have them complete & translated

fanciulli:
@volumio Aggiornato

PR #1770, opened 2019-06-11
Fix certain albums not playing and regression.

llevet:
Fix certain albums not playing
Fix the regression when attempting to play the album adds all tracks among albums of the same name to be added to the queue.


volumio:
Thanks I'll review it and ask people to test it and then merge. Well done

PR #1778, opened 2019-07-11
Feature/tunein

volumio:


spiralvoice:
It seems there is a bug when searching for stations using search phrases containing spaces:
https://forum.volumio.org/tunein-plugin-won-install-t13229-10.html

spiralvoice:
> It seems there is a bug when searching for stations using search phrases containing spaces:
> https://forum.volumio.org/tunein-plugin-won-install-t13229-10.html

Hopefully fixed by https://github.com/volumio/Volumio2/commit/0b9993603254ff470be7e081336e682aa4d5973a, test pending...

PR #1790, opened 2019-08-09
Minor improvement for logging error in label handling

bcornec:


bcornec:
I don't understand why you're closing this without applying the patch as this is a really simple fix, which just improves log files readability.

volumio:
Hi Bruno, simply because it had a conflict. If possible, please edit so we don't have a conflict and I'll happily merge it

PR #1796, opened 2019-09-07
Fix certain albums not playing after a Search

llevet:
Fix albums not playing (put on queue) after a search because the TAG albumartist is empty.

(bug report : https://forum.volumio.org/problem-playing-certain-albums-volumio-599-and-598-t12774.html)


volumio:
Thanks!

PR #1801, opened 2019-09-18
Update strings_sk.json

majko96:


volumio:
Thanks! Can you also update the UI language files? Thanks!
https://github.com/volumio/Volumio2-UI/blob/master/src/app/i18n/locale-sk.json

majko96:
You're welcome
Of course there is no problem :)

PR #1802, opened 2019-09-18
tracknumbers in albums view

gilphilbert:
Include the tracknumber of the current track in the return value so it can be provided by the API and WebSocket interfaces


volumio:
Well done!

Issue #1808, opened 2019-09-23
Websocket replaceAndPlay fails when index is 0


gilphilbert:
When you pass the websocket interface replaceAndPlay using the list and index pragma. The call fails and the queue is left empty. Passing any other index succeeds.

`io.emit('replaceAndPlay',{list: [{…}, {…}, {…}, {…}, {…}, {…}, {…}, {…}, {…}, {…}], index: 0})
`

I watched the logs (tail -f /var/log/volumio.log) and it seems that this request actually crashes the volumio service:
```
2019-09-23T20:44:36.528Z - info: CoreCommandRouter::volumioReplaceandPlayItems
2019-09-23T20:44:37.604Z - info: -------------------------------------------
2019-09-23T20:44:37.605Z - info: -----            Volumio2              ----
2019-09-23T20:44:37.605Z - info: -------------------------------------------
2019-09-23T20:44:37.606Z - info: -----          System startup          ----
2019-09-23T20:44:37.606Z - info: -------------------------------------------
2019-09-23T20:44:37.617Z - info: MYVOLUMIO Environment detected
2019-09-23T20:44:37.621Z - info: Plugin folders cleanup
2019-09-23T20:44:37.621Z - info: Scanning into folder /volumio/app/plugins/
```
(rest of boot process)

I know Javascript but I don't know much about debugging node environments, so I can't work out how to tell where the bug is.

xipmix:
You can get a long way with extra logging statements.
Looking at ```volumioReplaceandPlayItems``` it seems you should  hit this case:
```
    } else if (data.list && data.index) {                                       
        this.stateMachine.addQueueItems(data.list)                              
            .then(()=>{                                                         
                this.volumioPlay(data.index);                                   
                defer.resolve();                                                
            });                                                                 
```
It looks like the call to ```volumioPlay(data.item)``` doesn't happen, but a few more (temporary)```this.pushConsoleMessage``` could make the flow clearer. Also it can be useful to dump out the data structure to make sure it got to the function unscathed.

xipmix:
Hm, yes you should be seeing this log line
```
CoreStateMachine::play index 0
```
from ```CoreStateMachine.prototype.play```.

gilphilbert:
Yes, what's odd is I tried putting in some extra pushConsoleMessage lines and none of them printed. I've misplaced my dev machine (long story) but should have it back Monday night. I'll add more troubleshooting lines and see what I can find out about the data format.

[...]

PR #1810, opened 2019-09-27
Check 'uri' field is defined before using

xipmix:
It seems at least possible some data can be fed to this function that doesn't contain the 'uri' field.
Otherwise, why the check when calling ```self.commandRouter.logger.info``` ?


xipmix:
PS I noticed this while looking into issue #1808

xipmix:
Fair enough. I wasn't sure if there was some use case where promises without a uri were ok.

volumio:
Thanks!

PR #1812, opened 2019-10-01
Olchie master

volumio:


gebjohn:
Hi Volumio and Frank - what happened to this in the end? I can't see the feature in the latest release, so was it dropped? It's something I'd like, so if so, I'd happily pick up again and try and finish it. Volumio - can you point me in the right direction for what needs finishing up?

volumio:
@gebjohn As far as I remember, this PR was incomplete: it just listed composers but then the browsing was not executed

PR #1815, opened 2019-10-17
magic number phobia

xipmix:
untested...
apologies for the silly concatenation error (too much perl in my fingers)


volumio:
Even if it's a simple change, every PR should always be tested ;)

What is the reason for using a variable for max age?
Thanks ;)

xipmix:
Quite right. Tested now, on 2.657. No explosions, I don't see any regressions, albumart still loads. I am not sure how to check that the Cache-Control header is being output correctly, though.

My motivation for proposing this change is when I see raw numbers being repeated like this it makes me uncomfortable thanks to past experiences in other codebases - I find it to be a 'code smell'. I don't mind if you prefer not to merge it.

volumio:
@xipmix gotcha. I agree that it's cleaner and nicer. And I love when stuff does not explode ;)

PR #1818, opened 2019-10-19
[WIP] Use dependency injection for extra pluginmanager

piffio:
**This PR is still WIP, do not merge yet**

Make the stardard plugin manager myvolumio agnostic, by using dependency
injection to pass in the right "extra pluginmanager" when it
instanciated.
This will allow for passing alternative pluginmanagers for specific OEMs
as well as making the code more testable, which will become useful in
the future.


volumio:
@piffio is it still relevant? can you check against regression before merging?

PR #1825, opened 2019-10-27
Update config.json

balbuze:
Limited volume level for first start after a fresh install or factory reset


volumio:
Sorry @balbuze but I don't think this will be a good idea...

PR #1826, opened 2019-10-28
Replace and play fix

### LINKED ISSUE #1808: Websocket replaceAndPlay fails when index is 0

gilphilbert:
Fixes #1808 

Current code checks for the existence of data.index using:

```
} else if (data.list && data.index) {
```

Change to check for data.list!=undefined to avoid the situation where index.list=0 and the test fails when it should pass.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -479,7 +479,7 @@ CoreCommandRouter.prototype.replaceAndPlay = function (data) {
             		defer.resolve();
         	});
 		}
-    } else if (data.list && data.index) {
+    } else if (data.list!=undefined && data.index!=undefined) {
         this.stateMachine.addQueueItems(data.list)
             .then(()=>{
                 this.volumioPlay(data.index);
--- a/app/plugins/user_interface/websocket/index.js
+++ b/app/plugins/user_interface/websocket/index.js
@@ -104,12 +104,18 @@ function InterfaceWebUI(context) {
 
             });
 		
-			connWebSocket.on('addPlay', function (data) {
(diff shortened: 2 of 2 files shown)


gilphilbert:
Meant to open PR from a branch

PR #1827, opened 2019-10-28
Replace and play fix

### LINKED ISSUE #1808: Websocket replaceAndPlay fails when index is 0

gilphilbert:
Fixes #1808

Current code checks for the existence of data.index using:
```
} else if (data.list && data.index) {
```
Change to check for data.list!=undefined to avoid the situation where index.list=0 and the test fails when it should pass.


xipmix:
There is a very similar check a few lines below
```
    } else if ((!(data.list && data.index) && data.item && data.item.uri)) {
        this.stateMachine.addQueueItems(data.item)
```
should that be changed as well ? I suppose the question is

gilphilbert:
You're quite right - I've added a fix for that, too.

The crash seemed to be that the evaluation was false (since `if 0` evaluates to false) and this resulted in the wrong code being executed. The odd thing is that it should have been ending up here:
```
    } else {
        self.logger.error('Could not Replace and Play Item');
        defer.reject('Could not Replace and Play Item');
        }
```
...and there's no reason that should result in a crash.

gilphilbert:
Ah, yes there is! `self` isn't defined as `var self = this;` is missing from the beginning of this function. Added that to the PR.

xipmix:
Nice work. I wondered about the ```self``` thing, but missed the use of it in that ```else```.

xipmix:
One last thing I was thinking about. We had this structure
```
if (condition1) {
} else if (condition2) {
} else if (! condtion2 and condition3 and condition4) {
}
```
why can't we say:
```
if (condition1) {
} else if (condition2) {
} else if (condition3 and condition4) {
  // condition2 cannot be true
}
```
The way you've expressed ```! condition2``` makes a lot of sense too, though.

gilphilbert:
You are absolutely right - I was focusing on the smaller picture. There's no way to get to that line if those statements are true, so they're simply not required. I'm not sure why they were there, perhaps before lists of tracks were supported? Anyway, removed the extra evals  to make it more readable.

Thanks for checking over my code!

volumio:
Thanks!

PR #1863, opened 2019-12-20
volumio_command_line_client: add initrd edit

gkkpch:
This adds an option to edit the init script in an initramfs.
It will decompress the initrd and open nano as the editor.
After nano has exited, the initrd will be rebuilt.
Works for both known volumio initrd formats: volumio.initrd as built by the mkinitramfs-custom script and the imaged uInitrd version of it.

### DIFF
--- a/app/plugins/system_controller/volumio_command_line_client/commands/init-edit.sh
+++ b/app/plugins/system_controller/volumio_command_line_client/commands/init-edit.sh
@@ -1,0 +1,92 @@
+#!/bin/bash
+TMPWORK="/home/volumio/tmpwork"
+HELP() {
+  echo "
+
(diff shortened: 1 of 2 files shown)


xipmix:
Nice. Are the start/stopairplayplayback bits intended to be part of this PR?

volumio:
Merged with https://github.com/volumio/Volumio2/commit/45334b451eec1e2e3a942a028da7b730926f4485

gkkpch:
I discovered. a serious problem with uInitrd, it always re-creates an image in aarch64 format. we need to make this conditional for arm64 *and* arm.

ashthespy:
@gkkpch We also have that issue for our device template - should be maybe parse `lscpu`?

PR #1864, opened 2019-12-28
Make error message distinct

xipmix:
Another function, listRadioFavourites(), uses the same message
("Cannot list Favourites") which makes it difficult to distinguish
the source of an error if one occurs.


volumio:
Thanks! Maybe we should localize them?

xipmix:
Good idea, I'll take a look at that in another PR.

volumio:
Thanks

PR #1865, opened 2019-12-29
Filepath

xipmix:
It seems useful to include the path to the file that the function
failed to read in the error that is returned.

Untested, will test when I get a chance.


volumio:
Yes,please ping me when tested

xipmix:
Ok @volumio, no issue so far.

Tested by adding a song from music library to a new playlist, adding that to an empty queue and playing the track.

volumio:
Thanks!

PR #1869, opened 2019-12-31
Localise error messages

xipmix:
Tested by loading into 2.692 and restarting, no crashing etc was observed.

No idea how to force it to return an error, to exercise the changed lines.


volumio:
Thanks, very nice contributions from you ;)

Issue #1832, opened 2019-10-30
[2.657] Volume stuck at 100%, unable to to audio settings.


Akuba-:
After updating Volumio using the build in updater I have the problem that my audio settings where reset and I am unable to change them anymore. Volume will always be 100% and after I changed my settings it will not play anymore giving me an ALSA error with device hw0;0 not found.

While reading the log file I noticed this: `Cannot open shared library libasound_module_conf_pulse.so` and I guess this might be the cause of it.

Running Volumio on an Pi2.

Log: http://logs.volumio.org/volumio/q3cpGAY.html

jjjeroenski:
Same issue like @Akuba -, even with latest release 2.692
There is a fix if you click save submit on output devices after each reboot.

Using Rpi3b with Justboom Amp HAT
On previous releases never a issue

Log: http://logs.volumio.org/volumio/01XG3E2.html

xipmix:
When you say the volume is stuck at 100%, there are a few ways to interpret this.
 * Do you mean the volume-control-in-the-web-interface goes to 100% and you can no longer change it with the web interface controls?
 * Or, do you mean the sound output is maxed out and even though you can change the volume setting in the web interface, this makes no difference to the sound output level ?

@jjjeroenski - how does "save submit on output devices after each reboot" help improve things?

Separate question - does the problem happen with all music sources - music library, webradio, spotify ?

jjjeroenski:
@xipmix , i did a reboot just now and this is what i could see:
- notification sound when volume started doesn't work
- volume is at 0% even if in output options, there is startup volume 25.
- i can change manual volume from 0 tot 25 or 30 or ...
- i can start a song on usb storage or webradio, there is also sound, but i can't control sound from now. it' stuck. It's not the volume like you have changed before.

- if i open output options, i have output hardware : i2S dac on and justboom amp boards , i click "save" (notification player restarting) without changing something. --> volume control works again.
BUT it's more quiet than before, it's not the normal volume, but the volume control in the web gui up&down works again.

- you have to do this after each reboot and this is only since one of the latest versions.

i tested with usb storage, webradio and spotify, it's always the same after reboot and it's both less volume than before, for the same volume number.

Volume options:
-Mixer type: 'Hardware'
-Name mixer: 'digital'
Same like multiple posts in the volumio forum : https://forum.volumio.org/after-update-657-volume-control-t13107.html

xipmix:
Thanks for all that, it should help find the issue.

[...]

PR #1871, opened 2020-01-04
Selflogger

xipmix:
Use of 'this' looks inconsistent with the rest of the function but may be ok, I'll leave that up to you.
Tested on 2.692 pi. I don't have a hardware mixer but that shouldn't matter.

I was unable to get any debug-level log output from anywhere in volumio, so perhaps I'm missing something here.

Discovered while looking into #1832 .


volumio:
Even though this line is not used (debug is never used) it makes sense

PR #1872, opened 2020-01-04
Ensure checkValidMixer actually gets called

xipmix:
With 'self' not defined, the self.checkValidMixer condition is always false
and no mixers get added to the mixers array.

Fixes: #1832

Tested on a rpi with no Dac, no explosions.
Testing on an rpi with a Dac should be done soon - done, seems to fix the issue.


volumio:
Clever! This might fix the volume mixer issue!

Issue #1712, opened 2019-02-09
Airplay to volumio don't work


cmoulliard:
## Issue

When we redirect the sound output on a macbook to use `airplay`, then volumio UI displays the airplay symbol but we can't hear anything. Do I miss something ? Do we have to configure something ?

<img width="1432" alt="screenshot 2019-02-09 18 00 58" src="https://user-images.githubusercontent.com/463790/52523773-1e343d00-2c95-11e9-967d-378dd07775c0.png">

cmoulliard:
Can someone have a look please and tell me what is the problem ?

xipmix:
I can't find the log 'Airplay started streaming' in the code, the closest is

https://github.com/volumio/Volumio2/blob/2488f384538546350daead0ec9c4ab706450c79a/app/plugins/music_service/airplay_emulation/index.js#L225

Can you please tell what version you are finding this with, or perhaps just send a full log of a simple:
boot, airplay, sendlog sequence (https://volumio.github.io/docs/User_Manual/Sending_logs_for_troubleshooting.html)

cmoulliard:
here is the log : http://logs.volumio.org/volumio/u4xq76N.html
Version used: 2.668

xipmix:
Thanks
```
Jan 04 09:05:47 volumio volumio[1172]: info: Received update from a service different from the one supposed to be playing music. Skipping notification.Current webradio Received airplay_emulation
```
Just to clarify - did you have a webradio playing before trying to send an airplay stream to volumio? It's fine either way.

xipmix:
This is interesting, but not sure it's relevant yet. Throughout the whole time mpd.log is seeing this:
```
Jan 06 09:05 : player: Decoder is too slow; playing silence to avoid xrun
Jan 06 09:05 : player: Decoder is too slow; playing silence to avoid xrun
Jan 06 09:05 : player: Decoder is too slow; playing silence to avoid xrun
Jan 06 09:05 : player: Decoder is too slow; playing silence to avoid xrun
Jan 06 09:05 : player: Decoder is too slow; playing silence to avoid xrun
Jan 06 09:05 : player: Decoder is too slow; playing silence to avoid xrun
Jan 06 09:05 : player: Decoder is too slow; playing silence to avoid xrun
Jan 06 09:05 : player: Decoder is too slow; playing silence to avoid xrun
Jan 06 09:05 : player: Decoder is too slow; playing silence to avoid xrun
[...]

PR #1879, opened 2020-01-08
Fix: Add missing attribute of the alsa section in order to specify th…

cmoulliard:
- Fix: Add the missing attribute of the alsa section within the template of the shairport-sync template file  in order to specify the mixer to be used when the default is not used (e.g PCM)
- See : https://github.com/volumio/Volumio2/issues/1712


volumio:
Volume is already set by volumio

PR #1880, opened 2020-01-16
Fix next / prev functions while playing webradio (2/2)

TonyVpck:
This PR fix next / prev functions while playing webradio !

I tested the modifications suggested in this post below and the next / prev functions work correctly!
https://github.com/volumio/Volumio2/issues/1122

To fix the problem a first modification is necessary in 
Volumio2-UI/src/app/services/player.service.js

```
@@ -69,15 +69,11 @@ class PlayerService {
  }

  prev() {
-    if (this.state.trackType !== 'webradio') {
      this.socketService.emit('prev');
-    }
  }

  next() {
-   if (this.state.trackType !== 'webradio') {
      this.socketService.emit('next');
-    }
  }

  skipBackwards() {
```

Changes must be coupled with another from main volumio2-UI in 
Volumio2-UI/src/app/services/player.service.js

PR link : [volumio/Volumio2-UI/pull/711]

hoping it will be useful !

### DIFF
--- a/app/statemachine.js
+++ b/app/statemachine.js
@@ -1267,7 +1267,7 @@ CoreStateMachine.prototype.next = function (promisedResponse) {
         }
 	} else {
 		//self.setConsumeUpdateService(undefined);
-		if (this.isConsume && this.consumeState.service != undefined) {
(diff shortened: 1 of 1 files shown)


volumio:
Thanks!

PR #1883, opened 2020-01-19
Configure samba server-side file permissions

BenjaminDebeerst:
### Problem
The current volumio Samba configuration and setup has a number of issues.
* When mounted with cifs on a client, written files and directories show up with ownership nouser/nogroup on volumio
* Files and directories on an attached USB drive must be world-readable/-writeable to be readable/writeable from a samba client. The volumio user ownership should be sufficient.

### Root cause
* The volumio `smb.conf` does not set up any file/directory permissions for created files/dirs
* The volumio system does not set up any samba users at all ( `sudo pdbedit -L -v` returns no results on a vanilla volumio). Because guest access is allowed in `smb.conf`, all logins fall back to guest rights.

### Fix
* Amend `smb.conf` to contain file/directory creation permissions
* Setup a volumio user and password for samba

### Work in Progress
I've fixed the `smb.conf.tmpl` to use "sane" standard permissions (open for discussion). On my volumio, I set up the volumio user/password interactively and verified that it works.
For this PR. I've put the necessary command into the system controller plugin post-install hook, but I'm not clear whether this is the right spot in the first place, and uncertain about how to test this.

Let me know what you think.

### DIFF
--- a/app/plugins/system_controller/networkfs/smb.conf.tmpl
+++ b/app/plugins/system_controller/networkfs/smb.conf.tmpl
@@ -9,6 +9,10 @@ wins support = yes
 local master = no
 preferred master = no
 os level = 30
+create mask = 0664
+directory mask = 2755
+force create mode = 0644
+force directory mode = 2755
 
 [Internal Storage]
         comment = {NAME} Internal Music Folder


xipmix:
Nice to see this being addressed.  Will the smbpasswd step need modification if the volumio user password changes (see discussion #1192)? It probably should be the same as the user's system password.

BenjaminDebeerst:
It is not required that the samba password and the system password are the same. Samba maintains a separate password list. So technically, the answer to your question is "no".

I generally agree with the discussion in #1192 that changing the system password shouldn't break anything and with the security considerations. 

From a user perspective, it might make sense to keep the passwords in sync, although I don't think the Volumio system could enforce that? Most users changing the volumio default password via ssh would probably forget about samba, and I don't know of any way preventing that or reminding the user about this.

I'd propose the following:
1. Don't try to keep unix and smb passwords in sync. This keeps things simple and is in line with general system expectations.
1. Allow users to change both (independently). Changing passwords should not break standard functionality (the essence of #1192 for unix passwords).
1. Minimize Volumio code dependency on passwords.

For the second point: This is the case for samba right now, isn't it? I'm not aware of other places in the code referring to the samba password. That would be weird because there was no password set, yet.

To the third point: The samba user setup could be part of the system setup instead of the plugin code. Probably somewhere in the Build repository? With a pointer to the correct place to put some system setup code, I'm happy to move the user setup into a separate PR there.

I took a quick look, should the setup code go to any of these places?
https://github.com/volumio/Build/blob/master/scripts/volumioconfig.sh#L62
https://github.com/volumio/Build/blob/master/scripts/volumioconfig.sh#L121
https://github.com/volumio/Build/blob/master/scripts/configure.sh#L42

xipmix:
On reflection I think I agree with you about the password sync.
As to where to do the setup, my suggestion is in configure.sh#42, but there should be a comment in volumioconfig.sh#62 that there is a separate samba user with the same username but separate password.

BenjaminDebeerst:
Allright, I removed the user password setup from this PR and created volumio/Build/pull/400 as replacement. This makes the two PRs de facto independent.

BenjaminDebeerst:
Closing for staleness.

PR #1888, opened 2020-02-03
Add missing Intel ICH6 audio card, add audiophonics i-sabre DAC

gkkpch:


metapy123:
"Add audiophonics i-sabre DAC" was added to the Tinkerboard section making this change inaccessible to RaspberryPi users. It would be great to get this added for RasperberryPi as well.

Issue #1890, opened 2020-02-04
ALSA Card naming  inconsitencies 


ashthespy:
Was looking the logic of how we pick up the card's `id` and `name`  to match `cards.json` from: 
https://github.com/volumio/Volumio2/blob/e32e25d68d93383fe73e2a451c032b2661c71694/app/plugins/audio_interface/alsa_controller/index.js#L934-L942

This doesn't work some onboard codecs, such rockchips codecs, which seems to in the format `[rockchip,rkxxxx-acodec]` - currently such codecs will all get identified as `HDMI` because the `info` doesn't capture the entire parenthesis, defaulting `name` to `rockchip`.

Won't it be much more robust to replace all this splitting and replacing with a simple regex? 
Something along the lines of  `\^card\s(\d+).+?\[(.+?)\]\gm` would be sufficient right?

ashthespy:
Repo (but not old issues) has been moved over to https://github.com/volumio/volumio3-backend

PR #1891, opened 2020-02-04
Refactor and simplify `getAplayInfo` to be more robust

ashthespy:
Picking up from #1890, I've refactored `getAplayInfo()` to use regex to parse the card `id`, `name` and `device`, keeping the rest of the logic the same. Tested it on a few(two) boards, and seems to do the job. 

Some questions though:
1. Should `getAplayInfo()` also return `devices`?
1. Is there a reason why `namesArray` is needed? Are `TinkerAudio OnBoard` and `USB Audio OnBoard` devices on the same card?


volumio:
Sorry, I merged by mistake (did not test it). Could you please re open it?

PR #1902, opened 2020-02-24
WIP: Node 12 and Buster updates

ashthespy:
Branch for Buster + Node 12 packages, fixing small things as I go along!


ashthespy:
Ah, just noticed 9b62985 - and that you forked those repos for the fixes. Will add them here.
PS: 9b62985 has a syntax issue ;-)

ashthespy:
Closing this as it has gone quite stale..
Also, we should probably skip Node v12 and go with Node v14 directly. The built a few buster images with it and it was working fine..

PR #1906, opened 2020-02-29
WebRadio Update index.js to allow multiple space and accented characters in radio name

balbuze:
- allow search for radio with several space in name "france bleu pays de savoie"
- allow search for radio with accented character in name "france bleu isère" using https://www.unicode.org/reports/tr15/


volumio:
Thanks!

PR #1910, opened 2020-03-05
Refactor `getSystemVersion()` to be easily extendable

ashthespy:
This PR is two fold,

1. Refactor `getSystemVersion()` to be more extendable.
Also speed it up for while we are at it.. (Not super representative, most of the time would probably be spent in reading the file.. - but a test is a test ;-) https://www.measurethat.net/Benchmarks/ShowResult/97767)

2. Keep the parsed information in memory, and re-parse `/etc/os-release` only when required. 
  This is where it gets interesting, 
Currently `versionChangeDetect()` compares against `os-release` and saves `system_version` using `v-conf`. In this PR we still parse `os-release` when this function is called.
From a quick grep,  `versionChangeDetect()` is only called by a timeout function in [`startupSound()`](
https://github.com/volumio/Volumio2/blob/98dec1edae8efb6fce529b2a3d6065657d2e9e79/app/platformSpecific.js#L99-L102) 

The old `getSystemVersion()` that read the file each time on the other hand is a bit more popular (not counting all the times the WebUI calls for it)
```
logsubmit.js
app\index.js
app\plugins\music_service\webradio\index.js
app\plugins\system_controller\updater_comm\index.js
app\plugins\system_controller\system\index.js
app\plugins\user_interface\rest_api\index.js
app\plugins\user_interface\websocket\index.js
app\plugins\user_interface\rest_api\system.js
```
a. Can we assume that all these calls will be happy with the version from memory? 
b. I feel `versionChangeDetect` is redundant and can done in a less roundabout manor, by directly setting `system_version` during startup in `CoreCommandRouter`. Or am I missing something?

PS: There seems to a be quite some redundancy in `startupSound()` that can be moved into `CoreCommandRouter` instead.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -2056,10 +2056,42 @@ CoreCommandRouter.prototype.getPluginEnabled = function (category, pluginName) {
   return this.pluginManager.isEnabled(category, pluginName);
 };
 
-CoreCommandRouter.prototype.getSystemVersion = function () {
-  var self = this;
+// We assume that the system version information doesn't chnge during runtime,
+// so keep it in memory instead of parsing os-release each time.
+// TODO:
+//  Does the same hold for VOLUMIO_VARIANT?
+//  versionChangeDetect seems to handle this.
+//  by using config.set('system_version')
+CoreCommandRouter.prototype.getSystemVersion = function (force) {
+  // var self = this;
+  // return this.executeOnPlugin('system_controller', 'system', 'getSystemVersion', '');
+  const defer = libQ.defer();
 
-  return this.executeOnPlugin('system_controller', 'system', 'getSystemVersion', '');
+  if (force || !this.systemVersion) {
+    const osReleaseFile = fs.readFileSync('/etc/os-release').toString();
+    const releaseInfo = {
+      systemversion: 'VOLUMIO_VERSION',
+      builddate: 'VOLUMIO_BUILD_DATE',
+      variant: 'VOLUMIO_VARIANT',
+      hardware: 'VOLUMIO_HARDWARE'
+    };
+
+    for (const key in releaseInfo) {
+      const needle = releaseInfo[key];
+      const regEx = new RegExp(`^${needle}=(.*)$`, 'gm'); // overkill
+      const res = regEx.exec(osReleaseFile);
+      // No need to check if we actually captured something, will be null
+      if (res) {
+        releaseInfo[key] = res[1].replace(/^"|"$/g, '');
+      }
+    }
+
+    this.systemVersion = releaseInfo;
+    defer.resolve(releaseInfo);
+  } else {
+    defer.resolve(this.systemVersion);
+  }
+  return defer.promise;
 };
 
 CoreCommandRouter.prototype.getAdvancedSettingsStatus = function () {
--- a/app/index.js
+++ b/app/index.js
@@ -7,520 +7,504 @@ var execSync = require('child_process').execSync;
 var winston = require('winston');
 var vconf = require('v-conf');
 
-
 // Define the CoreCommandRouter class
 module.exports = CoreCommandRouter;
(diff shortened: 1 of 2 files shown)


volumio:
@ashthespy thanks!

2 comments:
- Please do not use a different linting in your PRs, otherwise its really hard to appreciate differences. Please redo the PR without linting differences.
- Please do not cache the result of os-release parsing: sometimes we do change os-release on the fly to test updates, and the overhead that results in reading the file each time is pretty neglectible.

ashthespy:
> Please do not use a different linting in your PRs, otherwise its really hard to appreciate differences. Please redo the PR without linting differences.

Ah, I thought the idea was to fix the whitespaces in each file as and when we update something in it, but either way kept the linting in a separate commit, so the actual changes are only in the last two commits :-) So it can be dropped before merge..  
But, could we bite the bullet and fix the white spaces in the entire repo once? The issue only gets worse as time passes! 

> Please do not cache the result of os-release parsing

The current logic is such:
1. Any call to `executeOnPlugin('system_controller', 'system', 'getSystemVersion', '')` will query 
`commandRouter.getSystemVersion()`, and add `additionalSVInfo` if present. 
 
2. `commandRouter.getSystemVersion()` parses `os-release` only the first time, and when a force flag is passed to it. 

3. `versionChangeDetect` calls `commandRouter.getSystemVersion(true)` to force a reparse. 


> : sometimes we do change os-release on the fly to test updates

Would it be fair to use `versionChangeDetect()` for the testing cases? 
Or we could set a `DEBUG_MODE` environment variable that forces a reparse for such testing cases? 

> and the overhead that results in reading the file each time is pretty neglectible.

There are quite a few front end interactions that also call this - plugin installation, system information, etc, i am not sure we have to parse os-release each time for such actions? 

PSS: The Code Inspector Ci doesn't seem to be set up properly..

volumio:
I will remove code inspector.

I agree on the proposed solution of force, it must be applied also when we check for new updates.
And yes, as soon as I find the time, I will lint everything with semistandard and we'll go with it

ashthespy:
Just saw https://github.com/volumio/Volumio2/commit/33642365f6069a578ebfc66f92fb4750e813e6a2, which got me curious - we seem to parse `/etc/os-relrease` directly in some places and use `getSystemVersion()` in others. 
Is there any particular reason for that? 
``` bash
# cat /etc/os-rlease 
# Stand alone stuff that don't hook into `commandRouter`.
# logsubmit.js # uses a local version of getSystemVersion()
# crashreport.js
# pluginhelper.js
# app\pluginmanager.js

app\plugins\audio_interface\alsa_controller\index.js
app\plugins\miscellanea\albumart\albumart.js
app\plugins\miscellanea\appearance\index.js
app\plugins\music_service\mpd\index.js
app\plugins\system_controller\updater_comm\index.js
app\plugins\system_controller\system\index.js

# getSystemVersion()
app\index.js
app\plugins\system_controller\updater_comm\index.js
app\plugins\system_controller\system\index.js
app\plugins\music_service\webradio\index.js
app\plugins\user_interface\websocket\index.js
app\plugins\user_interface\rest_api\index.js
app\plugins\user_interface\rest_api\system.js
```
For example `updater_comm` calls both [`getSystemVersion()`](https://github.com/volumio/Volumio2/blob/65cf1166679f02dccc50a29ce0ae12907e380787/app/plugins/system_controller/updater_comm/index.js#L300) and reads [`os-release`](https://github.com/volumio/Volumio2/blob/65cf1166679f02dccc50a29ce0ae12907e380787/app/plugins/system_controller/updater_comm/index.js#L251-L261)

I'd suggest breaking out this parsing logic in a single place, and use it from there. 
We could think of adding a module (say `sysInfo`) and then just import it in where ever it's needed. 

What do you think?

ashthespy:

PR #1911, opened 2020-03-05
Webradio : sorting the result in alphabetical oreder

balbuze:
With this PR, Radio stations are sorted in alphabetical order.
before result : 
old way sorted radio
after PR : 
new way sorted radio


volumio:
Excellent, great addition ;) small detail but very very nice. well done

PR #1912, opened 2020-03-06
Update strings_fr.json

balbuze:
add new fiels USER_INTERFACE


volumio:
Thanks! Could you please translate contemporary and classic as well?

balbuze:
I translated it, but as it is use as a name, it could have stayed like that ;-)

PR #1915, opened 2020-03-13
Execute plugin install scripts

### LINKED ISSUE #1627: Change plugin `install.sh` shell to bash?
Any reason for not using `bash` but `sh` to execute a plugins's `install.sh`?
https://github.com/volumio/Volumio2/blob/1e43e1e14289e4237ffdc30fc2579d650d530e64/app/pluginmanager.js#L1310-L1311

A quick grep shows that most of the plugin install scripts want to use bash:
```shell
1:#!/bin/bash
plugins/audio_interface/volsimpleequal/uninstall.sh
plugins/audio_interface/volsimpleequalmbeq/uninstall.sh
plugins/audio_interface/volsimpleequal/install.sh
plugins/audio_interface/volsimpleequalmbeq/install.sh
plugins/system_controller/volsmartpower/install.sh
plugins/system_controller/volsmartpower/uninstall.sh
plugins/audio_interface/volumiobauerfilter/uninstall.sh
plugins/miscellanea/volumiominidlna/install.sh
plugins/miscellanea/volumiominidlna/uninstall.sh
plugins/audio_interface/volumiobauerfilter/install.sh
plugins/music_service/volspotconnect/install.sh
plugins/music_service/volspotconnect/umountchroot.sh
plugins/music_service/volspotconnect/uninstall.sh
plugins/music_service/volspotconnect2/install.sh
plugins/music_service/volspotconnect2/onstart1.sh
plugins/music_service/volspotconnect2/uninstall.sh
plugins/audio_interface/volparametriceq/install.sh
plugins/audio_interface/volparametriceq/uninstall.sh
plugins/audio_interface/brutefir2/uninstall.sh
plugins/audio_interface/brutefir2/install.sh
plugins/music_service/spotify/install.sh
plugins/music_service/spotify/uninstall.sh
plugins/system_controller/wifireconnect/install.sh
plugins/system_controller/wifireconnect/uninstall.sh
plugins/audio_interface/volgrp/install.sh
plugins/audio_interface/volgrp/uninstall.sh
plugins/audio_interface/volstereo2mono/install.sh
plugins/audio_interface/volstereo2mono/uninstall.sh
plugins/miscellanea/playonconnect/install.sh
plugins/miscellanea/playonconnect/uninstall.sh
```

ashthespy:
Realised I never opened a PR for this..
Fixes #1627

PS: Dropped the whitespace fix commit this time ;)

### DIFF
--- a/app/pluginmanager.js
+++ b/app/pluginmanager.js
@@ -74,7 +74,7 @@ function PluginManager (ccommand, server) {
 
   var myVolumioPMPath = '/myvolumio/app/myvolumio-pluginmanager';
   if (fs.existsSync(myVolumioPMPath)) {
-    	this.logger.info('MYVOLUMIO Environment detected');
+    this.logger.info('MYVOLUMIO Environment detected');
     self.myVolumioPluginManager = new (require(myVolumioPMPath))(self.coreCommand, self.websocketServer, self.configManager, self.config);
   }
 }
@@ -179,7 +179,7 @@ PluginManager.prototype.loadCorePlugin = function (folder) {
       defer.resolve();
     }
   } else {
-	 	self.logger.info('Plugin ' + name + ' is not enabled');
+    self.logger.info('Plugin ' + name + ' is not enabled');
     defer.resolve();
   }
 
@@ -232,10 +232,10 @@ PluginManager.prototype.loadCorePlugins = function () {
 
   /*
     each plugin's onVolumioStart() is launched by priority order.
-	Note: there is no resolution strategy: each plugin completes
-	at it's own pace, and in whatever order.
-	Should completion order matter, a new promise strategy should be
-	implemented below (chain by boot-priority order, or else...)
+  Note: there is no resolution strategy: each plugin completes
+  at it's own pace, and in whatever order.
+  Should completion order matter, a new promise strategy should be
+  implemented below (chain by boot-priority order, or else...)
 */
   priority_array.forEach(function (plugin_array) {
     if (plugin_array != undefined) {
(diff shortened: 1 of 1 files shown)


ashthespy:
Can't suggest reviewers, so @xipmix tagging you for feedback..

ashthespy:
Thanks for the feedback!
> 
> Things I thought of
> 
>     * are we guaranteed to be able to chmod the file if there's no execute bit?

I would assume that it's best practice for the plugin dev to set the execute bit in the first place, and this is just a fall back. Eitherway, the entire plugin installer is wrapped with a `try...catch` so the dev should get feedback that it failed to install.

>     * are we assuming Bourne Shell (ash/dash, not bash) or allowing any kind of language to be used (there was discussion of this at one time, I forget the outcome)

Currently we assume that `install.sh` is a shell script, with a proper shebang set.
I wrote up a more generic implementation as well, that lets a `pluginScrip` be any executable file (py,perl etc), but don't want to push to many changes at once, so if this gets merged, and tested for a bit, will add that in :-)
 
>     * Might want to use a unique name for the /tmp file for each plugin installation

Have not changed any of that behaviour - AFIK, the install log gets overwritten on each install/uninstall action. That being said, the edge case of installing a plugin simultaneously is not handled (`volumio plugin install` + WebUI). But I would imagine that is going to fail else where first before reaching here..  

>     * in the GH viewer, indentation seems not to matching in the first hunk, the rest seems fine. I haven't checked with git tools

I haven't fixed the whitespace issues as per @volumio suggestions to keep changes easily reviewable, else they get lost in the white(space)noise ;-)

xipmix:
Thanks for the replies to my not-so-useful comments. I realised later that the plugin should get unpacked by the volumio user so should be owned by the same.. What you have looks sane to me.
I only mentioned the whitespace because the two hunks are inconsistently indented w.r.t. each other. I see no harm in adding a whitespace-only commit that fixes up lines already changed in the patch but that's just me I guess.

ashthespy:
Fixed the white space, thanks again for the comments! 
@volumio  From my side ready to test and merge :-)

ashthespy:
Rebased this on master to fix merge conflicts..

ashthespy:

PR #1917, opened 2020-03-15
Simple bench/metrics framework to time events

ashthespy:
##### Rationale:
We should have a way to keep track/time certain events, to gain insight into the system. As we add new features this becomes vital for user experience, and moreover gives the devs an idea on where to focus to improve things. 
Ideally, this data should also exposed to the `/dev` page and/or ultimately into the `downloader-v1` remote api. 

Not sure if this is the best way to do this, but it's a start (and requires no external dependencies..) 
I am also aware that given the async nature plugin manager, this doesn't represent true boot times - but that is a different matter..


volumio:
Good start ;)

PR #1918, opened 2020-03-15
Refactor timing out into `metrics` module

ashthespy:
Ah, didn't expect that quick a merge, was still cleaning up the implementation ;-) 
This should serve the same purpose, with a few added checks, and generally should be more scalable.

### DIFF
--- a/app/metrics.js
+++ b/app/metrics.js
@@ -1,0 +1,39 @@
+// A simple metrics/benchmarking framework
+
+class Metrics {
+  constructor () {
+    this.start = {};
+    this.end = {};
+
+    this.log = (label) => {
+      if (this.check(label)) {
+        this.end[label] = process.hrtime(this.start[label]);
+        console.log(`\u001b[34m [Metrics] \u001b[39m ${label}: \u001b[31m ${this.end[label][0]}s ${(this.end[label][1] / 1000000).toFixed(2)}ms \u001b[39m`);
+      } else {
+        throw Error(`Timer for ${label} not found`);
+      }
+    };
+
+    this.time = (label) => {
+      if (!this.check(label)) {
+        this.start[label] = process.hrtime();
+      } else {
+        throw Error(`Timer for ${label} already exists`);
+      }
+    };
+
+    this.check = (label) => {
+      if (typeof this.start[label] === 'undefined') {
+        return false;
+      } else {
+        return true;
+      }
+    };
+  }
+
+  dump () {
+    return this.end;
+  }
+}
+
+module.exports = Metrics;
--- a/app/pluginmanager.js
+++ b/app/pluginmanager.js
@@ -295,9 +295,9 @@ PluginManager.prototype.startCorePlugin = function (category, name) {
 	var plugin = self.getPlugin(category, name);
 
 	if(plugin)
-	{
+	{ 
 		if(plugin.onStart!==undefined)
-		{
+		{  
 		    var myPromise = plugin.onStart();
 			self.config.set(category + '.' + name + '.status', "STARTED");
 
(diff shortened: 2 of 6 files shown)


volumio:
Great that you improve ;) 
I'll review that tomorrow

ashthespy:
Okay, this should work for now. Added an entry to the /dev page:
image

ashthespy:
Added some more timing calls to the pluginmanger while working on another issue, but this isn't the true time to startup a plugin, as we don't really know when the plugin's `onStart()` actually finishes. 
image

ashthespy:
@volumio  Am I not hooking into the right place to get the plugin start up times? Cause it looks currently like the pluginmanager doesn't really resolve the `onStart` promise of each plugin properly. 
I initially tried hooking it in the `enable.then()` promise here:
https://github.com/volumio/Volumio2/blob/698e538caaf7c79578937ed5d4f3ec0e3dbe949b/app/pluginmanager.js#L1880-L1884

and also right after the `startPlugin` call https://github.com/volumio/Volumio2/blob/698e538caaf7c79578937ed5d4f3ec0e3dbe949b/app/pluginmanager.js#L1699-L1711

But they both don't give the actual values. For reference, I added some timing in the actual plugin's `onStart` after at the `defer.resolve()` and got more realistic values:
image

volumio:
Accurate timing depends on plugins and how they are written, some of them resolve the onstart promise immediately, some others after some actions has been performed.

Regarding the web ui, I find it an overshoot. What about a cmd in the style of systemctl blame?

ashthespy:
>Accurate timing depends on plugins and how they are written, some of them resolve the onstart promise immediately, some others after some actions has been performed. 

Not too sure about that -  I just made some tests with this snippet with the [`example_plugin`](https://github.com/ashthespy/Volumio-ExamplePlugin): 

```js
ControllerExamplePlugin.prototype.onStart = function() {
  var self = this;
  var defer = libQ.defer();
  const delay = (millis) => {
    return new Promise(resolve => setTimeout(resolve, millis));
  }
  const sleepfor = 6000; // Lets wait about 6 seconds
  const done = () => {
    // Once the Plugin has successfull started resolve the promise
    defer.resolve();
    metrics.log('Example');
  }
  // Start our timer
  metrics.time('Example');

  // Scenario A
  // setTimeout(done,sleepfor);

  // Scenario B 
  // delay(sleepfor).then(done);

  // Scenario C
  const sleepyQ = libQ.defer();
  setTimeout(() => {
    sleepyQ.resolve();
  }, sleepfor);
  sleepyQ.then(done);

  // Scenario Brute 
  // execSync("sleep 6"); // Put the thread to sleep. 
  // done()
  return defer.promise;
};
```
The first three strategies all result in `onStart` promise not being honoured, as `deffer.resolve()` is indeed set much latter. 
```
 [Metrics]  timing_plugin:  0s 0.65ms
 [Metrics]  Example:  6s 12.85ms
```

`Brute Scenario C` of course works, as it blocks the thread itself, making any `onStart.then()` moot:
```
 [Metrics]  Example:  6s 16.64ms
 [Metrics]  timing_plugin:  6s 17.44ms
```

Thought?

ashthespy:
From what I understand - no promises are resolved currently by the pluginmanger during boot, its just fire and forget:
https://github.com/volumio/Volumio2/blob/25647bd6aebf701ab307222139b3e122eebc19d2/app/pluginmanager.js#L398-L417

The only place the promises are resolved are in [`modifyPluginStatus()`](https://github.com/volumio/Volumio2/blob/25647bd6aebf701ab307222139b3e122eebc19d2/app/pluginmanager.js#L1393-L1404) and in `enableAndStartPlugin()`:
https://github.com/volumio/Volumio2/blob/25647bd6aebf701ab307222139b3e122eebc19d2/app/pluginmanager.js#L1699-L1711:

Is this by design?

ashthespy:
So managed to get more authentic times by resolving the promises on boot. And they seem quit consistent, making me believe that the bottle neck is the promise resolution strategy currently being used.. 




With regards to exposing this information on the dev page viz a command line option, while I do believe a lot more can be done with a CLI option (thresholds, sort, etc), I don't see to many people using it. With a simple section in the dev page, it's hidden away, but easily accessible metric for users to check what's slowing the system down. Also should help with the vague "long boot time" bug reports on the forum. That was my reasoning, happy to revert it if you don't agree..

ashthespy:

PR #1929, opened 2020-04-14
Fix airplay / shairport low volume output

hn:
Shairplay does not use the hardware mixer (at least with my external Raspberry Pi DAC) and as a result the volume is very low.

This fix inserts the 'mixer_control_name' setting in shairport.conf. Notably, the 'mixer' variable is exported in `airplay_emulation/index.js` (line 150) but not yet used in the config file.

Please include this fix -- or at least a commented line (`// mixer_control_name = ...`) in the config file to give a hint on how to set the hardware mixer.

### DIFF
--- a/app/plugins/music_service/airplay_emulation/shairport-sync.conf.tmpl
+++ b/app/plugins/music_service/airplay_emulation/shairport-sync.conf.tmpl
@@ -11,6 +11,7 @@ diagnostics =
 alsa =
 {
   output_device = "${device}";
+  mixer_control_name = "${mixer}";
   ${buffer_size_line}
 };
 


ashthespy:
IIRC there was duplicate of this that was closed for some reason..

hn:
I'm not an volumio/linux soundsystem expert. But for my setup the volume is too low without the 'mixer_control_name' statement (volume adjustment works in principle, but overall volume is extremely low when streaming music from an iPhone).

`//mixer_control_name = "${mixer}"; // necessary for some external DAC sound cards`

I suggest to include the commented statement so that others may have a hint for this solution. There are other comments within the file, so one single additional comment wouldn't decrease readability.

PR #1938, opened 2020-04-21
Refactor `volumioPlay` to combine `play` and `volatilePlay`

### LINKED ISSUE #1937: Discrepancy between `play` from WebUI and RestAPI/Websocket
Currently the WebUI tracks the currently playing service, and decides to call either `play` or `volatilePlay` from [`player.service.js`](https://github.com/volumio/Volumio2-UI/blob/53dec5e4a67006377c748c5a7ef44206c3c61a02/src/app/services/player.service.js#L46-L54) and [`player-buttons.html`](https://github.com/volumio/Volumio2-UI/blob/5dadb96add827e11380a48cfb8df55adb0cc5b7a/src/app/themes/volumio3/components/player-buttons/volumio3-player-buttons.html#L48-L49).

This breaks both the RestAPI and Websocket API's, as the user shouldn't be expected to keep track of the service, but should instead be incorporated into the statemachine. 

cc https://github.com/balbuze/volumio-plugins/issues/263

ashthespy:
Should fix #1937 and related upstream issues. 
The logic is if the current service is volatile and a track `index` is provided to `volumioPlay` we unset the volatile, and proceed to call `play`, other wise it just maps accordingly. 
We could then refactor the volatilePlay calls from the WebUI out at a later stage. 

PS: Not extensively tested (with myVolumio stuff),

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -1295,12 +1295,7 @@ CoreCommandRouter.prototype.explodeUriFromService = function (service, uri) {
 // Volumio Play
 CoreCommandRouter.prototype.volumioPlay = function (N) {
   this.pushConsoleMessage('CoreCommandRouter::volumioPlay');
-
-  this.stateMachine.unSetVolatile();
-
-  if (N === undefined) { return this.stateMachine.play(); } else {
-    return this.stateMachine.play(N);
-  }
+  return this.stateMachine.volumioPlay(N);
 };
 
 // Volumio Play
--- a/app/index.js
+++ b/app/index.js
@@ -1313,7 +1313,8 @@ CoreCommandRouter.prototype.volumioToggle = function () {
 
   if (state.status != undefined) {
     if (state.status === 'stop' || state.status === 'pause') {
-      return this.stateMachine.play();
+      // Use volumioPlay to account for volatile services
+      return this.stateMachine.volumioPlay();
     } else {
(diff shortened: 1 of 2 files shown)


ashthespy:
Rebased on master, and also fixed the Websocket and restAPI's `toggle` command as well.

johnk-:
Now fixed as tested here: https://github.com/balbuze/volumio-plugins/issues/263#issuecomment-626294548
> So far so good with the API and CLI with this version. Next, previous, stop, play and volume commands work perfectly. Toggle is still not working as expected. It stops, but doesn't continue. Toggle followed by a play command does work.

+ Toggle fix: https://github.com/balbuze/volumio-plugins/issues/263#issuecomment-626312811

I'll test the toggle today from both the API and the CLI and post the results here.

johnk-:
Toggle function tested and working from both REST API and CLI.

volumio:
We should test it with those plugins:
- Airplay
- Bluetooth
- Inputs 

If it breaks functionality, please report

johnk-:
Tested successfully with Airplay controlling iTunes via the API

illtrick:
@volumio would love to see this merged.  It's fixed a major use case for me and looks like a few others too.

ashthespy:

PR #1940, opened 2020-04-24
Update index.js

balbuze:
Do not not update DefaultMixer if output is set to Loopback (case for audio_interface plugins).

### DIFF
--- a/app/plugins/audio_interface/alsa_controller/index.js
+++ b/app/plugins/audio_interface/alsa_controller/index.js
@@ -664,8 +664,9 @@ ControllerAlsa.prototype.saveAlsaOptions = function (data) {
   }
 
   self.commandRouter.sharedVars.set('alsa.outputdevice', OutputDeviceNumber);
-  self.setDefaultMixer(OutputDeviceNumber);
-
+   if ((self.config.get('outputdevice')) != 'Loopback') {
+	self.setDefaultMixer(OutputDeviceNumber);
+   }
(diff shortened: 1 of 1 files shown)


balbuze:
it is much more complicated (for me) but I'll have look.
Just to know, what is the main advantage?

volumio:
I can take care of it, but worth explaining why:
- If you notice the setDefaultMixer gets called several times, but you're taking care only of one of them. so it will work only when someone sets the audio output, not on system restart
- It's better to add custom behaviours on the functions themselves, not on the part that calls them

balbuze:
ok,
but, in my plugin, Loopback is never saved in the config file.
- Volumio boot, hw output and mixer are set as if no plugin starts
- When the plugin start, it saves /data/configuration/audio_interface/alsa_controller/config.json and /boot/config.txt in /tmp (/boot/config.txt is saved because volumio disable i2s when Loopback is set.) - The plugin set the output to Loopback
- The plugin restores previously saved files
Ready for next boot!
It is complicated, but I had to compose with how Volumio actually works.

volumio:
Mmmm that's quite tricky then... Need to figure out something more...

balbuze:
I would say than 90% of the problem I met is due to this part... not from brutefir

volumio:
@balbuze problem is that rewriting the configuration files has a lot of uncontrolled consequences, that's why I think its a complex situation...

PR #1946, opened 2020-04-27
Cleanup language strings and other JSON files + Add CI checks for JSON

ashthespy:
This requires #1944 to be merged to avoid conflicts.

This checks for JSON errors:
1. In all the language strings in the `./app/i18n` directory: 
```bash
$ npm run $COMMAND
File: app/i18n/strings_de.json
Parse error on line 199, column 15:
...ert",    "DISABLED":"Deaktiviert",    ...
----------------------^
Duplicate key: "DISABLED"
```

2. In all `json` files in `./app/plugins` directory:
```bash
$ npm run $COMMAND
File: app/plugins/system_controller/services/config.json
Parse error on line 2, column 3:
{  ,  "plugin_name": "...
---^
Unexpected token ","
```


ashthespy:
@volumio Cleaned up all the duplicate keys, and sorted all keys alphabetically to make it more logical in the future.. 
From my side, this is ready to merge :-)

We could add a commit hook so that all contributions are also consistent. What do you think?

volumio:
@ashthespy great addition! 
For commit hooks: the issue is that husky works only if installed with npm install, and did not found a consistent way to have it on the images we build...

ashthespy:
We can ditch husky, and add simple hooks in `/.git/hooks` during image building?

volumio:
Yep I tried that, but didn't work... your help is appreciated on this.

May I ask one mod? Can you sort alphabetically only the first level keys? Otherwise it will be extremely difficult to spot diffs on the missing translated strings

ashthespy:
Aha, I see only now that there are whole categories missing from the strings. I assumed that all files had the same first level keys, and there were only second level missing keys.

ashthespy:
@volumio Added empty strings to all language files, and tweaked the `translateKeys` function to use the default language string, when the requested key is an empty string.

PS: Shutout to @balbuze  as strings_fr is the most up to date, and has no missing keys! ;-)

volumio:
Thanks! 
@balbuze you beat italians at their own game!

balbuze:
Lol😉

PR #1951, opened 2020-05-01
Remove dedicated logging to `/var/log/volumio.log`

ashthespy:
As we already log to console. 
Use `journalctl -o cat -u volumio.service` if required instead..


volumio:
Thanks

Issue #1952, opened 2020-05-01
volumio.log logs time in UTC instead of configured system timezone


danowar2k:
/var/log/volumio.log logs things in UTC, it seems. I've already run
`sudo dpkg-reconfigure tzdata`
and set the system timezone to my local timezone. Even after a reboot, the **volumio.log** still logs in UTC.

I've run node on the system. Typing
`d = new Date().toString()`
node shows the correct time. Looking around the internet, it seems node only logs the time in UTC when you (implicitly) use toISOString(). Node does that when you only type
`d = new Date()`
for example.

Can you fix the time in the log so that it shows the time in the configured timezone? (I hope that is only a change to a single line...)

danowar2k:
See
https://github.com/winstonjs/winston/issues/413#issuecomment-67208860

It seems Winston by default only logs in UTC unless you tell him to use another function as a timestamp

danowar2k:
Applied the path locally, didn't work...Hmm....

danowar2k:
Totally confusing, I just realized there aren't any github releases of Volumio. I have the current 2.729 which still uses winston 2.4.x . The master here uses winston 3.2.x (see package.json dependencies).

So I'm fumbling around trying to configure winston the 3.x way locally, but instead have to look for winston 2.x configuration...

danowar2k:
Okay, I have created a patch that works. The problem is, is there a branch that I can do a pull request for? The **master** branch here is not the one I have installed. I'm gonna check for another 2.x stable branch or something like that which still uses winston 2.x ...

danowar2k:
And so I just found out that volumio also logs to systemd (the console logger part) which can be read using `journalctl -u volumio` and already displays the correct time.

So we'll just call it quits here ;-)

PR #1953, opened 2020-05-01
Solution for #1952

danowar2k:
This lets winston log entries with the correct timezone it seems

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -15,7 +15,11 @@ function CoreCommandRouter (server) {
   this.logger = winston.createLogger({
     format: winston.format.simple(),
     transports: [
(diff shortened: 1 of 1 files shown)


danowar2k:
Don't merge this, this didn't work locally

danowar2k:
I have a new variant that works, but this shouldn't be merged with **master**. See the issue, I'll close this PR since I can't merge with master but need some other winston 2.x using branch...

volumio:
You happened to test in an unlucky moment... We updated package.json in sight of a future release.. On top of that, yesterday we disabled volumio.log facility... Using systemctl is the best way to go

PR #1962, opened 2020-05-12
Add a simple test for ALSA volume parsing Regex

ashthespy:
Follow up to #1961
We could refactor `getInfo` to use a single Regex and handle more cases. 

As a first step, here is a simple testing framework.  A good idea would be to collect multiple typical inputs, and also some edge cases.  Once we are happy with the regex, can refactor `getInfo` to use it.

```shell
volumio@volumio:/volumio$ node ./test/volume_controller.js

:::Input:::
 Simple mixer control 'SoftMaster',0
    Capabilities: volume
    Playback channels: Front Left - Front Right
    Capture channels: Front Left - Front Right
    Limits: 0 - 99
    Front Left: 10 [10%]
    Front Right: 10 [10%]
mute:  undefined
{ volume: 10, mute: false }

:::Input:::
 Simple mixer control 'PCM',0
  Capabilities: pvolume pswitch pswitch-joined
  Playback channels: Front Left - Front Right
  Limits: Playback 0 - 110
  Mono:
  Front Left: Playback 110 [100%] [0.00dB] [on]
  Front Right: Playback 110 [100%] [0.00dB] [on]

mute:  on
{ volume: 100, mute: true }
```

PS: *Unfortunately I can't get directly require and test with `volumecontroller` as it needs a proper framework with `commandRouter` and everything. It would amazing to have that, but don't have the time now to implement a full test harness...*

### DIFF
--- a/test/volume_controller.js
+++ b/test/volume_controller.js
@@ -1,0 +1,96 @@
+
+// const winston = require('winston');
+// const VolumeControl = require('../app/volumecontrol');
+const { execSync } = require('child_process');
+
+class VolumeControlTester {
+  constructor (context) {
+    // Regex to capture Volume [%] and possible mute status[on|off]
+    /*
+      /
+      \w+\s?\w+:(?:\s|\sPlayback\s)[-0-9]+\s\[([0-9]+)%\](?:[\n\r]|\s(?:[[0-9.-]+dB\]\s)?\[(on|off)\])
+      \w+
+      matches any word character (equal to [a-zA-Z0-9_])
+      \s?
+      matches any whitespace character (equal to [\r\n\t\f\v ])
+      \w+
+      matches any word character (equal to [a-zA-Z0-9_])
+      : matches the character : literally (case sensitive)
+      Non-capturing group (?:\s|\sPlayback\s)
+      Match a single character present in the list below [-0-9]+
+      \s matches any whitespace character (equal to [\r\n\t\f\v ])
+      \[ matches the character [ literally (case sensitive)
+      1st Capturing Group ([0-9]+)
+      % matches the character % literally (case sensitive)
+      \] matches the character ] literally (case sensitive)
+      Non-capturing group (?:[\n\r]|\s(?:[[0-9.-]+dB\]\s)?\[(on|off)\])
+      1st Alternative [\n\r]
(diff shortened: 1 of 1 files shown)


volumio:
wOW!

ashthespy:
Ah, that was a quick merge! Made some tweaks - b94360347171ea7ec01c09cccf0dd086625b4a22

xipmix:
Nice. I was trying to extend this but I am missing something basic about running the tests.
I want to use 'fs-extra' which is already being used by the language_files.js file, but it fails as well.
```
$ node ./test/language_files.js 
module.js:478
    throw err;
    ^

Error: Cannot find module 'fs-extra'
    at Function.Module._resolveFilename (module.js:476:15)
...etc...
```

ashthespy:
Are you running this on Volumio directly?  Else you have to install `fs-extra` as it's currently not listed as a `devDependency `

PS: I made some tweaks to this to allow us to directly test directly with`volumecontroler`. Will push that in a bit. 

PSS: The mute parsing should also test for `off` instead of `on`, if your testing with this branch..

xipmix:
I was in a git checkout of the backend, on my dev box. I'll have to install it, it would seem. The commit for language_files.js mentions mocha - have you worked out how to run that test (or yours) with mocha? I think we need a bit of extra doc here, which I'm happy to write.

ashthespy:
Take a look at my branch, there should be more functionality with the script. 
I did also manage to create a wrapper/mock test that can be run on the device to directly interact with the `volumecontroler` but it's still in a nascent stage, and don't have the time to put into it right now..

> The commit for language_files.js mentions mocha - have you worked out how to run that test (or yours) with mocha?

Tbh, I don't know much about the mocha test runner.

However, I do believe it's time we think of a proper test framework. 

Ideally we should have two test suites: 
- one that can be run on a dev machine, to test stuff like this simple parsing. 
- one to run on <a supported> device
I believe testing is is being done now manually by devs, but coverage is then quite specific to our use cases and devices. A test suite should catch such trivial errors and and guard against regressions. Not sure what @volumio's plans are about this though..

volumio:
I think that a testing suite will be excellent to have.
IMHO we should do tests for every new component from now on. 
As soon as Volumio3 framework proposal will be finalized, we'll include a specific strategy about testing.

PR #1966, opened 2020-05-14
alsa_controller index.js

balbuze:
Do not update mixer if output device is set to Loopback (by a plugin)


volumio:
Thanks! But, the difference in linting makes it very hard to understand what is changed. Can you please redoit in a way we can appreciate only what is changed?
Thanks!

Issue #741, opened 2016-10-05
Set Time and Date correctly using NTP service (timedatectl)


heis2201:
Looks like Volumio2 does not enable NTP by default.
Noticed that time was not set correctly, so had to run 
`timedatectl set-timezone EST` and then
`timedatectl set-ntp true`

Maybe we should include the first step as an input field in the GUI and the enable `ntp.service` by default

macmpi:
Would definitely be nice to have some UI in System pref panel to properly set date/time/timezone indeed, as great new timed features (Sleep and Alarm) are difficult to set without known time reference.
Thanks!

xipmix:
See also #373

heis2201:
Not a programmer here, so likely that I don't get the challenge described in #373 but

`timedatectl list-timezones > export-tz.txt` gives you a complete list of available timezones that can be used with `timedatectl set-timezone ...`

Why not have a select box with this entire list and upon *Save timezone* the selected string is send to little bash script that does `sudo timedatectl set-timezone selected_string`.

biva:
Hello, I tried the above operations, but I get an error: "Failed to set time zone: Access denied"
I used the password "Volumio".
So I can't change the timezone: what am I doing wrong?
Thanks!

macmpi:
how about `sudo dpkg-reconfigure tzdata`?

biva:
Great, that works! Thank you @macmpi !

macmpi:
but https://github.com/volumio/Volumio2/issues/741#issuecomment-256066520 is more systemd-friendly for setting this up in a future UI widget

PR #1971, opened 2020-05-18
Add methods to set/get devices time zone

### LINKED ISSUE #373: Set TimeZone
Currently there's no way to set the timezone via the web UI. This is a problem when using the Alarm functionality (which, granted, doesn't work terribly well right now #319).

It's not a big deal for me to run `dpkg-reconfigure tzdata`, but I imagine the console would be off-putting to some users.

---Alex

ashthespy:
Added some methods to set and get the current time zone.
~There is lot of scope of improvement, as `timedatectl` requires currently needs `sudo` perms.~
~If only I had read the docs, instead of skimming down to what I needed. [`--no-ask-password`](https://manpages.debian.org/buster/systemd/timedatectl.1.en.html)~ 
That also doesn't work, we just have to use the tried and tested `NOPASSWD`

<details>
<summary>  Roundabout method </summary>

We could set up `policykit` to allow the volumio user to use `timedatectl`, but I need to explore that more. Something along these lines *should* work..

```bash
$ sudo cat /etc/polkit-1/localauthority/10-vendor.d/10-time-policy.pkla
[Allow volumio user to set time zone]
Identity=unix-user:volumio
Action=org.freedesktop.timedate1.set-timezone
ResultActive=yes
```
</details>

On the FE side, we could pick up the current time zone from the browser in the wizard with 
[`Intl.DateTimeFormat().resolvedOptions().timeZone`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/resolvedOptions). Or, for a more wholesome experience, I am sure we can also find some angular lib for time zone selection..

Partially fixes #373 #741

PS: Also included preliminary tests that can be run on the device (with Volumio running)

### DIFF
--- a/app/plugins/system_controller/system/index.js
+++ b/app/plugins/system_controller/system/index.js
@@ -1074,7 +1074,7 @@ ControllerSystem.prototype.setTimeZone = function (data) {
   const defer = libQ.defer();
   const tz = data.name || 'UTC' // Fall back to UTC if nothing is provided
   this.logger.info(`Setting time zone to ${tz}`);
-  const tzcmd = 'echo volumio | sudo -S timedatectl set-timezone'
+  const tzcmd = 'timedatectl --no-ask-password set-timezone'
   exec(`${tzcmd} ${tz}`, (err, response) => {
     if (err) {
       this.logger.error(`Cannot set TimeZone: ${err}`);
--- a/app/plugins/system_controller/system/index.js
+++ b/app/plugins/system_controller/system/index.js
@@ -1074,7 +1074,7 @@ ControllerSystem.prototype.setTimeZone = function (data) {
   const defer = libQ.defer();
   const tz = data.name || 'UTC' // Fall back to UTC if nothing is provided
   this.logger.info(`Setting time zone to ${tz}`);
-  const tzcmd = 'timedatectl --no-ask-password set-timezone'
+  const tzcmd = 'sudo timedatectl set-timezone'
   exec(`${tzcmd} ${tz}`, (err, response) => {
     if (err) {
       this.logger.error(`Cannot set TimeZone: ${err}`);
--- a/app/plugins/system_controller/system/index.js
+++ b/app/plugins/system_controller/system/index.js
@@ -1046,3 +1046,41 @@ ControllerSystem.prototype.getCPUCoresNumber = function () {
       return 1
     }
 };
+
+ControllerSystem.prototype.getTimeZone = function () {
+  const defer = libQ.defer();
+  // Parse timedatectl's output
+  const timeZoneRegex = /(?:Time\szone):\s(\w+.\w+)\s\((\w+),\s([-+0-9]+)\)/gi
+  // Alternatively, we can just get the offset and acronym using 
+  // `date +"%Z %z"`
+  const tzcmd = 'timedatectl'
+  exec(tzcmd, (err, response) => {
+    if (err) {
+      this.logger.error(`Cannot get TimeZone: ${err}`);
+      return defer.resolve(err);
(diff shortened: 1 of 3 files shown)


ashthespy:
I was going about this all backwards, it just hit me we can just add `/usr/bin/timedateclt/` to [`sudoers.d/volumio-user`](https://github.com/volumio/Build/blob/b1bca506ec2c780c56506ef74fe004d0fe0a9eef/scripts/volumioconfig.sh#L123)

Ready for testing :-) 
```
$ node check_TimeZone.js
Emitting getTimeZone
Response pushTimeZone
 { name: 'Europe/London', acronym: 'BST', offset: '+0100' }
Emitting setTimeZone --> Europe/Rome
Response pushset TimeZone
success
Emitting setTimeZone --> CEST
Response pushset TimeZone
Error setting time zone: Error: Command failed: sudo timedatectl set-timezone CEST
Failed to set time zone: Invalid time zone 'CEST'

Exiting
```
Corresponding BE logs responses
```
info: CoreCommandRouter::executeOnPlugin: system , getTimeZone
info: setting timezone {"name":"Europe/Rome"}
info: CoreCommandRouter::executeOnPlugin: system , setTimeZone
info: Setting time zone to Europe/Rome
info: setting timezone {"name":"CEST"}
info: CoreCommandRouter::executeOnPlugin: system , setTimeZone
info: Setting time zone to CEST
error: Cannot set TimeZone: Error: Command failed: sudo timedatectl set-timezone CEST
Failed to set time zone: Invalid time zone 'CEST'
```

volumio:
Sorry, just saw it now.
What, in your opinion are the benefits of saving time zone?
Let me explain: our rationale for not having it is that a setting that is not striclty necessary. When we setup alarms, we automatically diff the browser time, so they are accurate without setting up the TZ.

What benefit would this bring? (considering we will have to change how alarm works)

ashthespy:
To be honest, it's just convenience/nice to have. I am sure a lot of users like me set this up manually right now on each instance.. 

Some SSL certificate errors pop up at times when the clock is not synced, but I don't see any core functionality issues of sticking to (ntp synced) UTC. 

Another advantage is supporting display plugins that have a clock while on standby. Else each plugin has to individually configure the time zone. ( E.g [`pydpiper.volumio.service`](https://github.com/dhrone/pydPiper/blob/6ec039436e94913e1b3c921f628e058ad64358d6/pydpiper.volumio.service#L29) )

Darmur:
it's useful for clock display on mdp_oled

volumio:
@Darmur I suggest not to use mpd oled: this will show only MPD metadata, and not all other services. 
If there is a library for node I can help to hook up the necessary functions

Darmur:
I'll have a look to mpd_oled, thanks for the suggestion.

BTW, I think even if it will be ported to node, proper timezone setting will be required to show the clock on display while not playing

volumio:
If we add this method, then we need to refactor the whole alarm plugin time calculation...

Darmur:
OK, I didn't look to the alarm plugin implementation to be honest. What I don't understand, if the plugin diff the browser time, why is it an issue to set a different time zone? In theory if you set your alarm after setting new time-zone and after a reboot, the delta should still be computed properly.

Darmur:
I can confirm changing time-zone doed not affect Alarm plugin. This is the test I did:

1) Clear all alarms
2) Set Amsterdam time zone from command line (sudo dpkg-reconfigure tzdata)
3) Reboot
4) Set some new alarms

It works fine, alarms were triggered right on time

ashthespy:

PR #1974, opened 2020-05-20
Add an entry in the dev page to tail logs

ashthespy:
Added a way to tail logs (via `journalctl -f`) to the dev page. This should make it easier for non console people to quickly get info that helps with debugging. There is a bit of overlap b/w this and the submit a log feature, but this can also help users debug stuff quickly. 
There is no max limit on the number of lines, but considering this is client side, should be fine?
Looks something like this:
image

### DIFF
--- a/app/plugins/system_controller/system/index.js
+++ b/app/plugins/system_controller/system/index.js
@@ -1053,7 +1053,7 @@ ControllerSystem.prototype.enableLiveLog = function (data) {
     try {
       this.logger.info('Launching a new LiveLog session');
       const format = 'cat'; // json is also an option for more serious logging/filtering
-      const args = ['/bin/journalctl','--output', format, '-f'];
+      const args = ['--output', format, '-f'];
       const defaults = {
         cwd: undefined,
         env: process.env
@@ -1066,7 +1066,7 @@ ControllerSystem.prototype.enableLiveLog = function (data) {
       if (this.livelogchild) {
         this.livelogchild.kill();
       }
-      this.livelogchild = spawn('/usr/bin/sudo',args, defaults);
+      this.livelogchild = spawn('/bin/journalctl',args, defaults);
 
       this.livelogchild.stdout.on('data', (d) => {
         liveLogData.message = d.toString();
--- a/app/plugins/system_controller/system/index.js
+++ b/app/plugins/system_controller/system/index.js
@@ -4,6 +4,7 @@ var libQ = require('kew');
 var fs = require('fs-extra');
 var config = new (require('v-conf'))();
 var execSync = require('child_process').execSync;
+const { spwan } = require('child_process');
 var exec = require('child_process').exec;
 var spawn = require('child_process').spawn;
 var crypto = require('crypto');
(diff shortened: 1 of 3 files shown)


app/plugins/system_controller/system/index.js
```diff
@@ -4,6 +4,7 @@ var libQ = require('kew');
 var fs = require('fs-extra');
 var config = new (require('v-conf'))();
 var execSync = require('child_process').execSync;
+const { spwan } = require('child_process');
```
volumio, line 7: While this is technically correct, it's inconsistent with current coding style. Please rephrase it as 
var spawn = require('child_process');

app/plugins/system_controller/system/index.js
```diff
@@ -1045,4 +1046,52 @@ ControllerSystem.prototype.getCPUCoresNumber = function () {
...
+      };
+      const liveLogData = {
+        message: 'Starting Live Log...\n'
+      };
+      this.commandRouter.broadcastMessage('LLogOpen', liveLogData);
+      
+      if (this.livelogchild) {
+        this.livelogchild.kill();
+      }
+      this.livelogchild = spawn('journalctl', args, defaults); // sudo or not?
```
volumio, line 1069: Should not require sudo, but we have an issue that in system updated from older installs the users file does not get updated properly. 
Another good practice is to use the full path.
I would do: /usr/bin/sudo /bin/journalctl -f

app/plugins/system_controller/system/index.js
```diff
@@ -1045,4 +1046,52 @@ ControllerSystem.prototype.getCPUCoresNumber = function () {
...
+        this.commandRouter.broadcastMessage('LLogProgress', liveLogData);
+      });
+
+      this.livelogchild.on('close', (code) => {
+        this.logger.info(`Live Log process terminated: ${code}`);
+        liveLogData.message = `process exited with code ${code}`;
+        this.commandRouter.broadcastMessage('LLogDone', liveLogData);
+      });
+    } catch (e) {
+      // The irony is deep.
```
volumio, line 1087: ahaahhaa

app/plugins/system_controller/system/index.js
ashthespy: Oops, sorry for the syntax, was copy pasting over from the plugin.. Will fix!

app/plugins/system_controller/system/index.js
ashthespy: fixed!

ashthespy:
Don't merge yet, there seems to be issue with sudo on older builds. Let me test more!

volumio:
Sorry, one of my suggestion was wrong, I get this error now:
May 20 16:59:27 studio3pi volumio[17428]: |||||||||||||||||||||||| WARNING: FATAL ERROR |||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
May 20 16:59:27 studio3pi volumio[17428]: { Error: spawn /usr/bin/sudo /bin/journalctl ENOENT
May 20 16:59:27 studio3pi volumio[17428]: at _errnoException (util.js:1022:11)
May 20 16:59:27 studio3pi volumio[17428]: at Process.ChildProcess._handle.onexit (internal/child_process.js:190:19)
May 20 16:59:27 studio3pi volumio[17428]: at onErrorNT (internal/child_process.js:372:16)
May 20 16:59:27 studio3pi volumio[17428]: at _combinedTickCallback (internal/process/next_tick.js:138:11)
May 20 16:59:27 studio3pi volumio[17428]: at process._tickCallback (internal/process/next_tick.js:180:9)
May 20 16:59:27 studio3pi volumio[17428]: code: 'ENOENT',
May 20 16:59:27 studio3pi volumio[17428]: errno: 'ENOENT',
May 20 16:59:27 studio3pi volumio[17428]: syscall: 'spawn /usr/bin/sudo /bin/journalctl',
May 20 16:59:27 studio3pi volumio[17428]: path: '/usr/bin/sudo /bin/journalctl',
May 20 16:59:27 studio3pi volumio[17428]: spawnargs: [ '--output', 'cat', '-f' ] }
May 20 16:59:27 studio3pi volumio[17428]: |||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||

I don't understand why spawn does not work this way, in any case the error should be handled

ashthespy:
I am not sure there is a way to catch errors from the parent spawn process, especially `ENOENT` errors. Normally you'd add a test to ensure the binary is present to avoid this, but since we don't dynamically change the command it should be fine? 

>Should not require sudo, but we have an issue that in system updated from older installs the users file does not get updated properly.
Another good practice is to use the full path.

I just noticed on an old version (2.5xx) requires sudo, while newer ones like you mentioned don't.. Anyway, have added it, and reworked it to use sudo and the full paths. But the `kill` doesn't work any more - as we are trying to kill `sudo`.. 

What to do?

volumio:
Good question... 
Anyway, since we deploy it on newer system, the use of sudo is not required, but we shall find a way to wrap it in a try catch

ashthespy:
Not sure the try catch is requires - with no `sudo`, `journnalctl` will only report "No Journal files found". It won't fail to launch..

PR #1987, opened 2020-06-30
Bugfix date metainfo not pushed

lbm-services:
Date metadata are not pushed.

### DIFF
--- a/app/plugins/music_service/mpd/index.js
+++ b/app/plugins/music_service/mpd/index.js
@@ -161,7 +161,8 @@ ControllerMpd.prototype.addPlayCue = function (data) {
       artist: cueItem.artist,
       album: cueItem.album,
       number: cueItem.number,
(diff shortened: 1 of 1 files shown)


xipmix:
Nice to see. 
What happens to sorting-by-date if the date is undefined, though? (e.g. an mpd query doesn't return that field) Should there be a default value (1970-01-01 or something?)

volumio:
Merged but changed date to year to harmonize the API layout

lbm-services:
mpd's tag_cache refers to it as 'Date''. I wanted it consistent this way.
Data could come from TDRC or TYER mp3-metatag. flac files store a DATE tag in comment block.
In rare cases freedb/musicbrainz return data in format YYYY-MM-DD but most players display only year anyway. 
A default value is unwanted. It should return an empty string.

Issue #1973, opened 2020-05-20
Raspberry Pi + Adafruit-MAX98357 - Audio output pops and clicks


timothyjward:
The Adafruit-MAX98357A is a cheap but flexible little I2S DAC/Amplifier, and really easy to set up with a Raspberry Pi Zero for a portable smart speaker. Volumio even has [built-in support](https://github.com/volumio/Volumio2/blob/a4e73ab3bab9527bb81808272661ed8bdf4b5586/app/plugins/system_controller/i2s_dacs/dacs.json#L3) for it.

The problem is that the Adafruit-MAX98357A isn't sent any I2S clock until some audio starts, then when the clock suddenly starts this causes a pop/click. This happens *every* time audio starts up, even after pausing a track.

The Volumio support is pretty minimal, and doesn't implement the [full setup suggestions for avoiding pops and clicks](https://learn.adafruit.com/adafruit-max98357-i2s-class-d-mono-amp/raspberry-pi-usage). 

As a user, what I want is for the built-in support to work without pops/clicks.

It is pretty simple to achieve, and I have successfully worked around it:

1. The audio must be routed through dmixer
2. There must be a system service playing from `dev/zero` into dmixer (e.g. using `aplay`)

This ensures that there is *always* an I2S clock (even if it's playing silence most of the time).

I have tried the following minimal `asound.conf`:

    pcm.softvolume {
        type             plug
        slave.pcm       "softvol"
    }

    pcm.softvol {
        type            softvol
        slave {
            pcm         "dmix:1,0"
        }
        control {
            name        "SoftMaster"
            card        1
            device      0
        }
    max_dB 0.0
    min_dB -50.0
    resolution 100
    }

With this service starting at startup:

    /usr/bin/aplay -D dmix:1,0 -t raw -r 48000 -c 2 -f S32_LE /dev/zero

I am aware that there are some limitations with this approach, the most significant of which is that most audio will end up being resampled (dmix runs at 48kHz 32bit by default). I see three reasons that this isn't a significant issue:

1. Audio has to be post-processed for this DAC anyway as it requires software volume control
2. Volumio has a high-quality resampled built in, so all audio can be sent to dmixer with the correct settings - this will avoid any low quality resampling lower in the stack.
3. The Adafruit-MAX98357 is really designed for driving relatively small low-power speakers, not super expensive reference quality drivers where you would be able to tell that resampling had occurred.

[...]

PR #1988, opened 2020-07-03
Allow plugins to contribute ALSA config

timothyjward:
This pull request makes a relatively simple but significant change to the ALSA config used in Volumio. 

Rather than having each audio component write directly to the audio hardware using `plughw:X,Y`, with special exceptions made in lots of places for `softvolume` when Software volume control is enabled, this solution always creates a PCM called `volumio` which is a `plug` intended to be used for audio output by everything.

The default ALSA config ends up looking like:

```
pcm.volumio {
    type        plug
    slave.pcm   volumioOut
}

pcm.volumioOut {
    type        hw
    card        X
    device      Y
}
```

This is identical in behaviour to `plughw:X,Y`, however it allows for plugins to contribute to the ALSA file in a simple, reliable way. For example the software volume control plugin can have the snippet:

```
pcm.softvolume {
    type        softvol
    slave {
        pcm   postVolume
    }
    control {
        name        "SoftMaster"
        card        X
    }
}
```

The overall generated ALSA config file ends up looking like:

```
pcm.volumio {
    type        plug
    slave.pcm   softVolume
}

pcm.softvolume {
    type        softvol
    slave {
        pcm   postVolume
    }
    control {
        name        "SoftMaster"
        card        X
    }
}

pcm.postVolume {
    type        plug
    slave.pcm   volumioOut
}

pcm.volumioOut {
    type        hw
    card        X
    device      Y
}
```

This has two significant advantages:

1. That *any* plugin can contribute to the ALSA configuration while still playing nicely with other plugins (e.g. The equaliser plugin doesn't work with software volume control)
2. That all audio inputs (e.g. shairport-sync) will reliably work with the expected audio configuration as they point at `volumio` not directly at the hardware

This PR makes it much simpler to fix issue #1973 - a "dmix volumio plugin" would be trivial to create which adds in a dmix pcm to the volumio chain, and an "on/off" slider can enable/disable the background silent audio. (I'm planning to work on a plugin for this next).

This PR also makes it much simpler to provide things like working SnapCast integration. The current plugin relies on "patching" the ALSA and MPD configuration in ways that don't work reliably (re-patching buttons have been added). Any time a configuration change touches the ALSA config then SnapCast breaks.

I've done my best to verify that this works in all directions (to/from softvolume, to/from dac, volume changes etc). Let me know if I can do anything more to help with it.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -3,6 +3,7 @@
 var libQ = require('kew');
 var libFast = require('fast.js');
 var fs = require('fs-extra');
+var exec = require('child_process').exec;
 var execSync = require('child_process').execSync;
 var winston = require('winston');
 var vconf = require('v-conf');
@@ -42,7 +43,7 @@ function CoreCommandRouter (server) {
   this.pluginManager.pluginFolderCleanup();
   this.configManager = new (require(__dirname + '/configManager.js'))(this.logger);
 
-  this.pluginManager.startPlugins();
+  var pluginPromise = this.pluginManager.startPlugins();
 
   this.loadI18nStrings();
   this.musicLibrary.updateBrowseSourcesLang();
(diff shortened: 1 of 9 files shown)


volumio:
Thanks TIM!
This is exactly the direction we wanted to go to provide streamlined integration with alsa plugins and a new snapcast implementation. 
Since this will be an important step forward, I would love to discuss about that in PVT if you like.
If yes, would you mind writing to michelangelo at volumio dot org?

PR #2000, opened 2020-09-15
Change order of strings to be the same as the source strings

ordtrogen:
In preparation for completing the translation, I have edited strings_sv.json so that the order of the translatable strings is the same as in strings_en.json. Then when translating, you can work with them side-by-side and it's easier to keep track of where you are in the process.

I have made no changes to the actual strings, even though several corrections are needed. I wanted to avoid editing any strings as the reordering would have made it difficult to see what had been changed. Will work on that after merge


ashthespy:
My fuzzy memory vaguely recalls some plumbing to have this done automatically.. 
Probably a good idea to setup a CI check manage these chores.

ordtrogen:
If you look a bit at the various strings_XX.json files, e.g strings_ca.json, strings_da.json, strings_es.json etc, you can see how strings that do have a translation have sort of bubbled up to the top of each group, with all the empty translations trailing at the end. And this breaks the alphabetical sorting.

Since this seems to be a pattern across several files, my guess is that some automatic step in the process has created this situation, unless translators have been instructed to order the string like this, which instruction I'm not aware of.

Could it be related to the jsonlint command you mention?

ashthespy:
> Since this seems to be a pattern across several files, my guess is that some automatic step in the process has created this situation, unless translators have been instructed to order the string like this, which instruction I'm not aware of.

Actually, I think it's the opposite - I think it's just a copy-paste artefact.. 

>My fuzzy memory vaguely recalls some plumbing to have this done automatically..

I found it: #1950 but it doesn't seem to have been merged. 

@volumio Doesn't this make adding new strings less work?

ashthespy:
@ordtrogen #2007 should have sorted all language keys now :-)

ordtrogen:
Looks great! Thanks.

BTW, @volumio , shouldn't something similar be done for the project Volumio2-UI? The file app/i18n/locale-sv.json in that project has a lot fewer strings than the corresponding locale-en.json

I guess I should close this PR?

PR #2003, opened 2020-09-26
AudioInjector.net iso. soundcard needs module loaded with old kernels

flatmax:
The audio injector isolated soundcard requires the snd_soc_cs4271_i2c 
module to be manually loaded with old 4.x.x kernels.


flatmax:
Autoload module for old Raspberry Pi Kernels.

volumio:
@flatmax 

Couple of comments:
- Did you test it, does it work?
- To make it future proof, for when we will update the kernel (no module loading necessary), what about using a script that checks if kernel < 5.x (with uname -a) then loads the kernel?

flatmax:
The procedure for testing isn't clear.
Can you please guide on testing this ?

I can't imagine it would matter if it were attempted to load in a kernel which has already loaded it ?

volumio:
To test: simply make the mod on your system. Disable i2s, reboot, enable i2s with this DAC and see if that works (also after a reboot).

volumio:
@flatmax do you have any news on the testing?

flatmax:
I haven't had a chance to work on Volumio unfortunately.

PR #2006, opened 2020-10-07
[pluginhelper] Add `--production` flag to npm install

ashthespy:
Should slim down `node_modules` for plugin that use `devDependencies`


volumio:
Really Clever! Thanks Ash! We will also use it for packaging Volumio's node modules

PR #2018, opened 2020-11-08
Update dacs.json

balbuze:
add HifiBerry amp2 (HifiBerry Dac + overlay : https://www.hifiberry.com/shop/boards/hifiberry-amp2/ )
Any need for : https://github.com/volumio/Volumio2/blob/master/app/plugins/audio_interface/alsa_controller/cards.json ?


volumio:
Thanks!

PR #2030, opened 2020-12-15
cards.json: add missing 'HDA Intel'

gkkpch:


gkkpch:
This card is (still) used in a number of Intel NUC devices

PR #2033, opened 2020-12-28
Add all items in explodeUri when adding to playlist

fanciulli:
Go through all items of the explodeUri instead of limiting to the first item


volumio:
@fanciulli for what is it for?

PR #2035, opened 2021-01-04
Preparing for node14

ashthespy:
Taking the opportunity to clean-up/harmonise the ~black hole~node_modules

#### Upgrade packages for Node v14
- [ ] `inotify` - needs git version 
- [x] `libxmljs` -- `^0.18.4 -> ^0.19.7`
- [x] `mdns` --   `^2.3.3 -> ^2.5.1`
- [ ] `udev` - needs git version

#### Unused packages 
I found these packages with a rather quick approach

```
───────┬──────────────────────────────────────────────────────────────────────
       │ File: .\utils\checkUnused.sh
───────┼──────────────────────────────────────────────────────────────────────
   1   │ #!/usr/bin/env bash
   2   │ # Quick script to find unused node_modules
   3   │ # Given our repo structure `depcheck` and friends won't work
   4   │ # Consider using smarter regex?
   5   │ # /\w+\s(\w+)\s=\srequire\(\'(\w+)\'\)/g
   6   │
   7   │ echo "Checking for un(required) node_modules"
   8   │ for row in $(jq -r '.dependencies | keys | .[]' package.json); do
   9   │   if [[ ${#row} -gt 1 ]]; then
  10   │     pkg=${row}
  11   │     echo "Checking for <$pkg>"
  12   │     if ! git grep -q "require('${pkg}')"; then
  13   │       echo "${pkg}" is not used!
  14   │       unused+=("${pkg}")
  15   │     fi
  16   │   fi
  17   │ done
  18   │ echo -e "\n\nUnused modules:\n${unused[*]}"
  19   │ echo "Check again for false positives!"
  20   │ # ejs firebase mqtt node-cache onoff require-from-string timsort
───────┴──────────────────────────────────────────────────────────────────────
```
It's not smart enough to catch for example [`ejs`](https://github.com/volumio/Volumio2/blob/c9e4849495e810bd6ebb6833232daba25d86dd4c/http/index.js#L36). 
It is also not smart enough to catch stuff like `yargs` that is required but never used. But `eslint` should catch such things as seen here.. https://github.com/volumio/Volumio2/blob/52aad0e2315ceb3b7f95677722214c4c72195a14/crashreport.js#L4


volumio:
Please add back the myvolumio modules as you suggested ;) Until this is sorted can't merge...
Please also have a look at https://github.com/volumio/Volumio2/tree/node14modules

ashthespy:
Closing this in favour of the unified node_modules approach..

PR #2041, opened 2021-01-14
Send empty array if no urls are defined

a-i-ks:
Sending back an empty arrary string is better than sending an error string in the body of the response. 

When writing a client application, one assumes that a string array of URLs will be sent as a response. If instead a custom message is sent as a response when no URLs have been entered, this message must be handled separately in the client, which is extremely unpleasant.


volumio:
Correct ;)

PR #2046, opened 2021-01-19
fix: play from genre not working

janbar:
Hi, here a fix for play an album or artist from genre not working.


janbar:
I pushed force to fix an other issue in the original code:
`var genreString = curUri.replace('genres://');`
This line store "undefined" in place of "genres://" breaking the parse later. I replaced it by:
`var genreString = curUri.replace('genres://', '');`

crieke:
Thanks for your PR, janbar.
It is highly appreciated.

PR #2058, opened 2021-01-28
Plugin dependency check

timothyjward:
Perform an `npm` style check using `engines` to determine whether a plugin can run on the current Volumio

### DIFF
--- a/app/i18n/strings_en.json
+++ b/app/i18n/strings_en.json
@@ -312,6 +312,7 @@
         "REMOVE_SUCCESS": "Removed"
     },
     "PLUGINS": {
+        "CHECKING_DEPENDENCIES": "Checking that the plugin is suitable for this version of Volumio",
         "CHECKING_DUPLICATE_PLUGIN": "Checking for duplicate plugin",
         "CONFIRM_PLUGIN_INSTALL": "Install Plugin",
         "CONFIRM_PLUGIN_INSTALL_WARNING_MESSAGE": "Warning, community plugins are developed by third party developers and might result in unexpected behaviours and system instabilitiy. Are you sure you want to install this plugin?",
--- a/app/pluginmanager.js
+++ b/app/pluginmanager.js
@@ -1162,6 +1162,33 @@ PluginManager.prototype.checkPluginDependencies = function (folder) {
   return pendingResult.then(() => { return result });
 };
 
+PluginManager.prototype.listPluginsBrokenByNewVersion = function (newVolumioVersion) {
+  var self = this;
+  var result = [];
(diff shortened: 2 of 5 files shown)


volumio:
Really well done! 
Just one note: would there be a quick way to avoid installing semver module (and using a function instead?)

timothyjward:
> would there be a quick way to avoid installing semver module (and using a function instead?)

I know that the `node_modules` directory is a big part of the volumio image and needs to be managed carefully. The semver module is small (~60kB) and has a small dependency chain (2 modules @ 15kB each). This feature actually use quite a lot of semver's function, and handling versions and version ranges gracefully is actually pretty difficult. I don't think we could produce something reliable very easily, particularly if we wanted to catch the edge cases around the use of `~`, `^`, `.x`, `*` etc.

ashthespy:
@timothyjward  We (@balbuze + me) seem to be having some [issues with this](https://github.com/balbuze/volumio-plugins/pull/299#issuecomment-799642924)..

```
Checking that the plugin is suitable for this version of Volumio Volumio version null not usable with plugin. 
The plugin cannot be installed on this version of Volumio
```

It looks like `semver` doest like our Volumio version number, because well, it isn't Semver!  The leading `0` in the second `minor` part is not to spec. 

```js
const semver = require('semver')
volVers = '3.052'
volSemver = '3.52'
const VVersion = semver.coerce(volVers);
const VSVersion = semver.coerce(volSemver);
console.log(VVersion);
console.log(VSVersion);
# 

null
SemVer {
  options: {},
  loose: false,
  includePrerelease: false,
  raw: '3.52.0',
  major: 3,
  minor: 52,
  patch: 0,
  prerelease: [],
  build: [],
  version: '3.52.0'
}
```

timothyjward:
> It looks like semver doest like our Volumio version number, because well, it isn't Semver! The leading 0 in the second minor part is not to spec.

That does seem to be the issue. Using loose matching should be sufficient as `3.052` becoming `3.52.0` should be fine. It will still be lower than `3.100`, which will become `3.100.0`, so the absolute ordering isn’t violated.

ashthespy:
Yep but currently it parses `null` and doesn't let you install any plug in with volumio in the engines section..

timothyjward:
I’m just not near enough a keyboard to do a fix today, but adding `, {loose: true}` into the coerce calls in this PR should be enough to fix it.

volumio:
When possible, can you update it please?

volumio:
Great! 
@ashthespy @balbuze can you confirm it works so we merge it?

PR #2060, opened 2021-01-30
Playlist optimization

fanciulli:
When playing a local playlist queues the first item and starts playback. The remaining items are then queued. This should introduce a lot better performances for long playlists with items from services llike Qobuz and Tidal


volumio:
@fanciulli 

This code does not solve anything at all. If a TIDAL playlist is played, it's not even invoked. It refers to playlists of files only.

IMPORTANT. For next times:
- Only submit PRs if appropriately tested with the actual behaviour
- Testing must be careful and accurate

PR #2062, opened 2021-02-01
More cleaning up of node_modules

ashthespy:
Bumped a few, removed others..

### DIFF
--- a/app/playlistManager.js
+++ b/app/playlistManager.js
@@ -2,7 +2,6 @@
 
 var libQ = require('kew');
 var fs = require('fs-extra');
-var S = require('string');
 var singleBrowse = false;
 
 module.exports = PlaylistManager;
@@ -689,14 +688,8 @@ PlaylistManager.prototype.commonPlayPlaylist = function (folder, name) {
 
           var uris = [];
           for (var i in data) {
-            var uri;
-            var fullUri = S(data[i].uri);
-
-            if (fullUri.startsWith('music-library')) {
-              uri = fullUri.chompLeft('music-library/').s;
-            } /* else if (fullUri.startsWith('/')) {
-							uri = fullUri.chompLeft('/').s;
-						} */ else uri = data[i].uri;
+            // Remove leading `music-service/` identifier if present
+            var uri = data[i].uri.replace(/^music-library\//, '');
 
             var service;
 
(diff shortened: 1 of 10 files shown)


volumio:
We need yargs for the plugin helper.
if modclean is not installed we shall remember to install it globally before packing modules

ashthespy:
There are still a lot of low-hanging fruit for the picking... 
Example `underscore` 
`_.isEmpty(filters)`  -> `!filters`
https://github.com/volumio/Volumio2/blob/52aad0e2315ceb3b7f95677722214c4c72195a14/app/musiclibrary.js#L362-L366
Looking at the code, it's not very clear what is being done, but you could also just use a `try/catch` if you want to just check if data is valid JSON or already an object.. 

`_.each()` -> `Array.prototype.map()`
https://github.com/volumio/Volumio2/blob/52aad0e2315ceb3b7f95677722214c4c72195a14/app/plugins/system_controller/network/lib/index.js#L169

`_.extend()` -> `Object.assign()`
https://github.com/volumio/Volumio2/blob/52aad0e2315ceb3b7f95677722214c4c72195a14/app/plugins/system_controller/network/lib/index.js#L60

IMO, the entire vendored  [`node-wireless`](https://github.com/tlhunter/node-wireless) added to [`network/lib/index.js`](https://github.com/volumio/Volumio2/blob/52aad0e2315ceb3b7f95677722214c4c72195a14/app/plugins/system_controller/network/lib/index.js) should be reevaluated.. 

And ironically -- https://github.com/volumio/Volumio2/blob/52aad0e2315ceb3b7f95677722214c4c72195a14/app/randomqueue.js#L3-L4

ashthespy:
> We need yargs for the plugin helper.

Hmm where? Don't see any require statement for it in [`pluginhelper.js`](https://github.com/volumio/Volumio2/blob/52aad0e2315ceb3b7f95677722214c4c72195a14/pluginhelper.js)

volumio:
Can we "reduce the scope" of this?
IMHO:
- Bumping packages is ok 
- Remove redundant `body-parser` middleware is ok
- Remove redundant `String` module is ok

Can you undo the rest of those?

volumio:
@ashthespy sorry but I am having troubles testing it.
IMHO next times:
- One PR for each change
- Having modclean removed breaks the package bundling, when removing something lets make sure current procedures still work

ashthespy:
Ah, this flew under my radar -- feel free to cherry pick, will leave the branch up for now. 


> *  Having modclean removed breaks the package bundling, when removing something lets make sure current procedures still work

Ah wow, didn't realise you were still running this bundle script from a running Volumio on the pi directly!

PR #2074, opened 2021-02-10
Fix/default bg selection

volumio:
Fixes setting random background on first boot

### DIFF
--- a/app/plugins/miscellanea/appearance/index.js
+++ b/app/plugins/miscellanea/appearance/index.js
@@ -189,7 +189,7 @@ volumioAppearance.prototype.getBackgrounds = function () {
 
   fs.readdir(backgroundPath, function (err, files) {
     if (err) {
-      console.log(err);
+      self.logger.error('Failed to read Background files: ' + err);
     }
     files.forEach(function (f) {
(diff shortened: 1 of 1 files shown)


app/plugins/miscellanea/appearance/index.js
```diff
@@ -273,7 +277,15 @@ volumioAppearance.prototype.createThumbnailPath = function () {
           if(pluginFiles.every(function(f) { return dataFiles.includes(f); })) {
             return null;
           } else {
-            return libQ.nfcall(fs.copy, __dirname + '/backgrounds', backgroundPath);
+            fs.copy(__dirname + '/backgrounds', backgroundPath, function(err) {
```
timothyjward, line 280: This change does a few things I'm not sure about:
1. It moves the default background selection to before the thumbnail generation (previously it was afterwards, although I'm not sure it matters)
2. It will always run the thumbnail generation, even if the copy operation fails.
3. It does not wait for the result of setting the default background, which is asynchronous (this may have been a pre-existing problem).

volumio: 1) does not matter
2) that should not be an issue
3) that is not a problem as well

PR #2076, opened 2021-02-10
Update the plugin manager to better handle MyVolumio plugins

timothyjward:
Report enabled for enabled MyVolumio plugins
Return the plugin folder for MyVolumio plugins

### DIFF
--- a/app/pluginmanager.js
+++ b/app/pluginmanager.js
@@ -346,6 +346,12 @@ PluginManager.prototype.getPackageJson = function (folder) {
 PluginManager.prototype.isEnabled = function (category, pluginName) {
   var self = this;
 
+  // MyVolumio plugins aren't very good at updating the enabled flag - just check that 
+  // the plugin is valid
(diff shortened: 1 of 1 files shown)


app/pluginmanager.js
```diff
@@ -346,6 +346,12 @@ PluginManager.prototype.getPackageJson = function (folder) {
 PluginManager.prototype.isEnabled = function (category, pluginName) {
   var self = this;
 
+  // MyVolumio plugins aren't very good at updating the enabled flag - just check that 
+  // the plugin is valid
+  if(self.myVolumioPluginManager.checkIfPluginIsInCurrentPlan(category, pluginName)) {
```
ashthespy, line 351: You should check if `myVolumioPluginManager` is not `undefined` -- else will fail for all non myvolumio (community) build devices that don't contain myVolumio stuff.. :-)

volumio: Yep, that's the culprit

PR #2079, opened 2021-02-19
Adding telemetry to mpd command issuing

volumio:


ashthespy:
Are you sure this measures what you expect it to measure?
https://github.com/volumio/Volumio2/blob/e3bac8408d618d0bf0143cc7504ece3b3fe8e9ed/app/plugins/music_service/mpd/index.js#L323-L346
 I am assuming you actually want to measure the time taken for the resolved promise, and not just to initiate the promise?

PR #2083, opened 2021-02-27
Fix `Volume.vol` type when parsing Software mixer

ashthespy:
Was a `String`, should be a `Number` as per rest of the code in this module..

### DIFF
--- a/app/volumecontrol.js
+++ b/app/volumecontrol.js
@@ -497,8 +497,8 @@ CoreVolumeController.prototype.retrievevolume = function () {
           self.logger.error('Cannot read softvolume: ' + error);
         } else {
           var volume = stdout.replace('\n', '');
-          currentvolume = volume;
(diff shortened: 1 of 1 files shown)


volumio:
@ashthespy did you test it?
This is the result of parsing stdout so I don't recall if that's a number or string

ashthespy:
stdout is a string, plus there is the `String.replace()` just above in `var volume = stdout.replace('\n', '');` 

Tested it for a Software mixer, but I didn't look very deep into what else this fixes.
For example this check would have failed previously, defaulting to `100` always for Software mixer. 
https://github.com/volumio/Volumio2/blob/bbb4d290de421e20eae41b45cc85bb96c28960fa/app/index.js#L170-L178

volumio:
Gotcha, did not notice the parseInt

PR #2085, opened 2021-02-28
Fix and refactor device HWUUID creation

ashthespy:
There are a multiple places in the codebase where a device ID is created using the hash of the mac address. However this is currently being done in multiple places with different strategies. The one thing in common is that they all assume that there is a `eth0` network interface present. 

For example in `system` there is [`getHwuuid()`](https://github.com/volumio/Volumio2/blob/5dba14aa04ca954c0d7d63d6f0f623cfdefb32c0/app/plugins/system_controller/system/index.js#L1011-L1023)
https://github.com/volumio/Volumio2/blob/5dba14aa04ca954c0d7d63d6f0f623cfdefb32c0/app/plugins/system_controller/system/index.js#L1015-L1023
This is repeated inside the same `system` again in [`callHome()`](https://github.com/volumio/Volumio2/blob/5dba14aa04ca954c0d7d63d6f0f623cfdefb32c0/app/plugins/system_controller/system/index.js#L566-L576)
https://github.com/volumio/Volumio2/blob/5dba14aa04ca954c0d7d63d6f0f623cfdefb32c0/app/plugins/system_controller/system/index.js#L566-L576

Apart from the fact that this would ignore devices with no `eth0`, the replacement would basically convert `00:0a:95:9d:68:16` -> `000a:95:9d:68:16` as `String.replace()` will replace only the first occurrence of `:` as it's not a Regex. 

This could all be done rather robustly using the NodeJS's inbuilt [`os.networkInterfaces()`](https://nodejs.org/docs/latest/api/os.html#os_os_networkinterfaces), that would work even if a device doesn't have `eth0` (say a device with only WiFi) or when we remove move away from the legacy [naming scheme to predictable iface names](https://wiki.debian.org/NetworkInterfaceNames#THE_.22PREDICTABLE_NAMES.22_SCHEME)

### DIFF
--- a/app/plugins/system_controller/system/index.js
+++ b/app/plugins/system_controller/system/index.js
@@ -578,7 +578,7 @@ ControllerSystem.prototype.callHome = function () {
   info.then(function (infos) {
     if ((infos.variant) && (infos.systemversion) && (infos.hardware) && (hwuuid)) {
       self.logger.info('Volumio Calling Home...');
-      exec('/usr/bin/curl -X POST --data-binary "device=' + infos.hardware + '&variante=' + infos.variant + '&version=' + infos.systemversion + '&uuid=' + hwuuid + '" http://updates.volumio.org/downloader-v1/track-device',
+      exec('/usr/bin/curl -X POST --data-binary "device=' + infos.hardware + '&variante=' + infos.variant + '&version=' + infos.systemversion + '&uuid=' + hwuuid.uuid + '" http://updates.volumio.org/downloader-v1/track-device',
         function (error, stdout, stderr) {
           if (error !== null) {
             if (calltrials < 3) {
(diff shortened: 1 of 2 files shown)


ashthespy:
Oh and then there is another method employed in `updater_comm` by [`pushUpdatesSubscribe()`](https://github.com/volumio/Volumio2/blob/a2bc69f6ab9381fbb9acb1b4422d111a5676e9a1/app/plugins/system_controller/updater_comm/index.js#L243-L252)
https://github.com/volumio/Volumio2/blob/a2bc69f6ab9381fbb9acb1b4422d111a5676e9a1/app/plugins/system_controller/updater_comm/index.js#L243-L252

This hash would be different as it doesn't strip out the first `:` like done in `system`

We should really consider the moving things around into modules to improve code reusability and reduce the amount of repeated code in this codebase!

volumio:
Good start. 

While at it can you also refactor https://github.com/volumio/Volumio2/blob/a2bc69f6ab9381fbb9acb1b4422d111a5676e9a1/app/plugins/system_controller/updater_comm/index.js ?

(use executeOnPlugin to call the existing funcion in system)

ashthespy:
This is breaking change - so not sure how you'd like to handle it - should I just shim the external API's from these two below to return only `uuid` and not `isHW`? 
Or is this being called elsewhere (myVolumio?) directly via `executeOnPlugin` ?

`CoreCommandRouter.prototype.getHwuuid()`
https://github.com/volumio/Volumio2/blob/df871fa4816701362f7add98178af058833d9fff/app/index.js#L2151-L2155

and `getDeviceHWUUID` 
https://github.com/volumio/Volumio2/blob/df871fa4816701362f7add98178af058833d9fff/app/plugins/user_interface/websocket/index.js#L1738-L1743

volumio:
How is it breaking change? Can we make it in a way it is retro-compatible?

ashthespy:
Breaking change in two ways.

1. For the same MAC, the new hash is different. 
  -- Previously it was `00:0a:95:9d:68:16` -> `000a:95:9d:68:16` -> md5Hash
  -- Now it is `00:0a:95:9d:68:16` -> `000a959d6816` -> md5Hash
  So all previous device hashes will now no longer match for the same MAC.

2.  We add the `isHW`  field that `pushUpdatesSubscribe`  requires to `getHwuuid`
```js
// previously
getHwuuid() -> <String>
// breaking
getHwuuid() -> { uuid: <String>, isHW: <Boolean> } 
```

As this affects only a small selection of devices (Rpi Zero, NUCs  with only WiFi), perhaps we should give up and leave it for some other day?

volumio:
@ashthespy thanks.

For next times: the rule is always do retro-compatible stuff and avoid breaking changes.
I suggest to:
- Send back string as hwuuid, always
- Let's not bother the isHW in updaterComm (as it's not useful anyway) and evaluate it always to true (this is not used).

ashthespy:
Ah, this slipped under my radar..
Way was trying to make this codebase a bit more [DRYish](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).  But I don't see an elegant solution that is non breaking, so going to give this a pass.

It is not a real blocker/big issue anyway, so we can always fix it later, once the server back-end gets fixed..

Issue #501, opened 2016-07-01
[RC2] Can't play anything


arnodeclercq:
Can't seem to play anything with RC2 on Raspberry Pi 2 B.
I can add a folder, it scans and adds everything to library. But can't play/start any file or stream.
If I try a stream it gets added to the queue but just shows the link. Files don't even get added to the queue.
I don't have a DAC, I just use the jack output. Didn't change any settings.
If you need more information let me know.

faead:
Same problem.
But with Raspberry 3 and Hifiberry DAC+.
When I try to add a playlist to the queue I hear the startup sound. Directly added files crush the whole webinterface.

Fresh installed system (2x).

PR #2087, opened 2021-03-01
Fix installToDisk for x86 devices (old & new)

gkkpch:


gkkpch:
This PR also needs [Build PR #501](https://github.com/volumio/Build/pull/501)

volumio:
Thanks!

PR #2093, opened 2021-03-16
Handle cases when Volumio's version isn't valid semver

timothyjward:


timothyjward:
Do not merge this until either @ashthespy or @balbuze have verified that it fixes the problem discussed in #2058

app/pluginmanager.js
```diff
@@ -1105,67 +1105,70 @@ PluginManager.prototype.unzipPackage = function () {
...
         if(package_json.engines.volumio) {
-            volumioCheck = true;
+            result.volumioCheck = true;
             return self.coreCommand.getSystemVersion()
                 .then(e => {
-                    var volumioVersion = semver.coerce(e.systemversion);
-                    if(!semver.satisfies(volumioVersion, package_json.engines.volumio)) {
-                        message += 'Volumio version ' + volumioVersion + ' not usable with plugin. ';
+                    var volumioVersion = semver.coerce(e.systemversion, { loose: true });
+                    if(nodeVersion === null) {
```
ashthespy, line 1135: Should be `volumioVersion === null` no?

timothyjward: Good spot - I was attempting to work quickly and clearly missed a copy/paste. I've pushed another commit which should fix this, although the `{loose: true}` should make the null a very rare occurrence.

ashthespy:
Can't test on device for a few more weeks, so will leave it to @balbuze to test in the meantime :-)

PR #2094, opened 2021-03-16
added codec and bitrate to radio stations in browse

crieke:
When the info is available, codec and bitrates are displayed in radio browser. 
This is not always accurate and propagates the info from the used radio directories.

### DIFF
--- a/app/plugins/music_service/webradio/index.js
+++ b/app/plugins/music_service/webradio/index.js
@@ -360,14 +360,14 @@ ControllerWebradio.prototype.listRadioForGenres = function (curUri) {
             var name = children[i].attr('name').value();
             var id = children[i].attr('id').value();
             var bitrate = children[i].attr('br').value();
-
+            var streamcodec = children[i].attr('mt').value();
 
             var category = {
               service: 'webradio',
               type: 'webradio',
               title: name,
-              artist: bitrate + ' kB/s',
-              album: ' ',
+              artist: self.formatCodecString(streamcodec) + ' (' + bitrate + ' kbps)',
+              album: '',
               icon: 'fa fa-microphone',
               uri: 'http://yp.shoutcast.com' + '/sbin/tunein-station.m3u' + '?id=' + id
             };
@@ -435,13 +435,14 @@ ControllerWebradio.prototype.listTop500Radios = function (curUri) {
             var name = children[i].attr('name').value();
             var id = children[i].attr('id').value();
             var bitrate = children[i].attr('br').value();
+            var streamcodec = children[i].attr('mt').value();
 
             var category = {
               service: 'webradio',
               type: 'webradio',
               title: name,
-              artist: bitrate + ' kB/s',
-              album: ' ',
+              artist: self.formatCodecString(streamcodec) + ' (' + bitrate + ' kbps)',
+              album: '',
               uri: 'http://yp.shoutcast.com' + base + '?id=' + id
             };
             try {
@@ -964,13 +965,15 @@ ControllerWebradio.prototype.searchWithShoutcast = function (search) {
             var name = children[i].attr('name').value();
             var id = children[i].attr('id').value();
             var bitrate = children[i].attr('br').value();
+            var streamcodec = children[i].attr('mt').value();
+
 
             var category = {
               service: 'webradio',
               type: 'webradio',
               title: name,
-              artist: bitrate + ' kB/s',
-              album: ' ',
+              artist: self.formatCodecString(streamcodec) + ' (' + bitrate + ' kbps)',
+              album: '',
               uri: 'http://yp.shoutcast.com' + base + '?id=' + id
             };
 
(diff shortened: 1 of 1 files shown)


app/plugins/music_service/webradio/index.js
```diff
@@ -359,12 +359,14 @@ ControllerWebradio.prototype.listRadioForGenres = function (curUri) {
...
             var id = children[i].attr('id').value();
+            var bitrate = children[i].attr('br').value();
+            var streamcodec = children[i].attr('mt').value();
 
             var category = {
               service: 'webradio',
               type: 'webradio',
               title: name,
-              artist: '',
+              artist: self.formatCodecString(streamcodec) + ' (' + bitrate + ' kbps)',
```
ashthespy, line 369: You could add a `bitrate` filed and fix the logic over at [`statemachine.js`](https://github.com/volumio/Volumio2/blob/a40b108519dbb2023427ac77fb2ed10a4c0e7cbb/app/statemachine.js#L112-L117) so it actually gets used?

xipmix: I was also a bit surprised to see bitrates going into the 'artist' field. I see it's just continuing the current (ab)use of the field, so fair enough if this is deemed urgent I guess. The statemachine.js code uses different terms btw - 'samplerate' and 'bitdepth'. It would be nice to have them consistent but that's for a separate PR.

xipmix: I looked into this a bit more. Mostly the code uses 'samplerate' and 'bitdepth' (one assumes bitdepth=samplerate*bitdepth, or maybe one has to multiply by the number of channels as well?).
If a stream only supplies a bitrate and not the other two, then presumably it's not possible to figure those values out?
Unless there's a convention I'm not aware of that allows assuming a bitdepth or samplerate...

Below are some brief notes from a bit of grep & less
```
app/plugins/miscellanea/wizard/wizard_steps/conf/spotify.json                   
  spotify uses the bitrate term                                                 
                                                                                
app/plugins/music_service/mpd/index.js                                          
                                                                                
  ControllerMpd.prototype.parseTrackInfo = function (objTrackInfo)              
    returns mpd track info                                                      
    it uses the bitrate term if the term exists in the argument object          
    not necessarily only in the streaming case                                  
    It doesn't check for samplerate or bitdepth, or set them.                   
                                                                                
  ControllerMpd.prototype.parseState = function (objState)                      
    returns a json associative array with a bitrate field,                      
      if the argument object has a bitrate field it assumes Kbps.               
    It doesn't check for samplerate or bitdepth, or set them.                   
                                                                                
app/plugins/user_interface/mpdemulation/helper.js                               
  sets a dummy bitrate of 0                                                     
  does this weird thing where it makes a string like so                         
  status.audio = message.samplerate + ':' + message.bitdepth + ':' + message.      channels
  the comment says the default is 44000:24:2                                    
                                                                                
app/statemachine.js                                                             
  CoreStateMachine.prototype.getState = function ()                             
    it looks at self to see if it's in 'consume' state                          
      if so and the current object is a webradio                                
        it sets samplerate and bitdepth fields to ''                            
                                                                                
  CoreStateMachine.prototype.syncState = function (stateService, sService)      
    does similar things, but only in play mode?                                 
                                                                                
  we have objects that have all three fields                                    
  strangely getState does not always return an object that has bitrate,         
  but it seems to always return samplerate & bitdepth                           
                                                                                
                                                                                
app/plugins/audio_interface/alsa_controller/UIConfig.json                       
  samplerate not used as a word, grep picked it up as a substring of               target_samplerate etc
                                                                                
app/plugins/audio_interface/alsa_controller/index.js                            
  ditto                                                                         
                                                                                
app/plugins/music_service/airplay_emulation/index.js                            
  a few functions that return an object with samplerate & bitdepth but not bitrate
                                                                                
app/playqueue.js                                                                
  the objects here always get a samplerate & bitdepth, never a bitrate          
```

timothyjward: Apologies if this is all obvious, but I think the confusion about bitrate is that it is defined by the file and encoding, not the audio output.

* Sample Rate: The number of audio samples per second. 
* Bit depth: The number of bits used to encode each sample in the *output* audio stream - unless you're playing raw/pcm audio this will be *bigger* than the number of bits per sample in the input audio stream.
* Channels: The number of channels in the audio. 
* Bit Rate: This is the data rate for the incoming audio stream. This may be something like 128kbps for MP3 web radio, or 700+kbps for a CD quality FLAC file. This gives an idea of the quality of the incoming audio. The output audio bit rate would be *much* higher, typically 1.4Mbps for 44100:16:2 audio, but this isn't usually an interesting number

I think it's therefore necessary to provide sample rate, bit depth, channel count, and bit rate. They all mean different things and may be helpful to the user. Normally you should be able to query the stream for all four things. The bit rate will likely vary throughout the input stream, but the sample rate, output bit depth and channel count *should* remain constant.

volumio:
Merging. As Tim pointed rightfully out, this is different info.
It can be easily later refactored when we add the relevant info into the UI

PR #2100, opened 2021-03-27
a-z lists

joehajek:
integrate a-z lists in volumio - useful for big libraries
.) for artists/genres/albums possibility to start with a a-z list instead of a huge list of artists/genres/albums
.) sorting of artists/genres/albums with optional ignoring articles (The Beatles are sorted under letter B instead of T)
.) optionally ignoring artists/albums/genres with fewer than a given nr of tracks in the lists
.) all options configurable in the UI
.) language files only german and english


joehajek:
grafik
grafik
grafik

PR #2103, opened 2021-04-07
Merge Buster/dev

ashthespy:
Seems to have gotten left behind. 
Initially testing images were built against this staging branch, and release versions were against `buster/master`. But if this staging branch is no longer being used, we might as well remove it and open PRs directly to `buster/master`?

PS: Perhaps squash and merge to avoid the extra merge commit..


volumio:
Yep, until released we better do PR on buster/master

Thanks and welcome back!

PR #2104, opened 2021-04-07
Temporarily disable scanning internal (non-removable) disks

gkkpch:
This disables scanning of internal disks to prevent a deadlock at first boot.
It will be replaced by a more appropriate solution where the use should be able to select the disks to add to the library, much like adding a remote share.
USB disks are not affected


volumio:
This needs a slight refactoring but in pricinple it's good ;)

PR #2106, opened 2021-04-09
Do not scan internal HDDs

volumio:


gkkpch:
```
 if (device.ID_BUS === 'usb') {
```
This part does not do the same as my original ref #2104
```
var fsRemovable = execSync('/bin/lsblk -rno tran "/dev/$(lsblk -rno PKNAME ' + device.DEVNAME + ')"',{ uid: 1000, gid: 1000, encoding: 'utf8'});
if (fsRemovable.indexOf('usb') >= 0) 
```
The problem is that device.ID_BUS does not always report the original bus, see [https://community.volumio.org/t/volumio-3-buster-beta-for-rpi/45263/220](url)
The solution from #2104 does.

PR #2116, opened 2021-05-06
Fix issue where moving first item of the queue is not possible

Joni-Salminen:
We should not check if value is greater than 0, since then we cant move first item in the queue.


Joni-Salminen:
As all brainwork is done at nights, there should be checks in place for negative values as well, or we will introduce new possible bug, i sort that out when get from work.

Joni-Salminen:
Should be ok now

xipmix:
Tested-By: xipmix@users.noreply.github.com

This fixes a real problem - if you want to move the first item in the queue down, the item moves in the UI but then snaps back to the first position. Any other item in the queue can be moved. Tested with v2.861, classic UI and 3 web radios in the queue.

It would be nice to see the error message at line 1561 updated as well
```
- this.logger.error('Cannot move item in queue, from or to parameter missing');
+ this.logger.error('Cannot move item in queue, from or to parameter missing, undefined or < 0');
```

PR #2117, opened 2021-05-06
Disable repeatSingle when repeat is disabled

Joni-Salminen:
repeat single does not work alone, since its nested inside repeat checks and coupled with it in other checks, so it should be set false when repeat is disabled. Currently "pushState" can emit "repeatSingle == true", but if repeat is false, we are not repeating single track, so the state is wrong.

I believe this is more of correcting the state sent to listeners, real fix would be to allow repeatSingle actual repeat the single song when its set, i can look into that if its more preferred way?

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -1555,7 +1555,7 @@ CoreCommandRouter.prototype.volumioMoveQueue = function (from, to) {
   var defer = libQ.defer();
   this.pushConsoleMessage('CoreCommandRouter::volumioMoveQueue');
 
-  if (from !== undefined && to !== undefined) {
+  if (from && to) {
     return this.stateMachine.moveQueueItem(from, to);
   } else {
     this.logger.error('Cannot move item in queue, from or to parameter missing');


xipmix:
Looks nice and simple but it prompts questions:
What version is this occurring on?
Can you describe a simple way to reproduce what you are seeing? 
It would help me understand this issue better.
 - How many queue items do I need? 
 - I'm not quite sure how to 'repeat single' in the queue UI either, there is only a control to repeat the whole queue (v2.861)

Also for a short series like this it helps clarity if you can rebase before pushing to your fork, or just cherry-pick the changes that were actually needed into a new branch. I often do a 'git diff master -- <modified>.js' and use that to work out the final series of commits I need.

Joni-Salminen:
It's occuring on latest official release 2.882.

I don't think it's possible to reproduce it via the webcontrol UI, i did find this out when developing native android client/websocket library for Android, if you try to enable "repeatSingle" without enabling "repeat" with the command "setRepeat", then the pushState event emits "repeatSingle:true" but it's not actually enabled.

PR #2126, opened 2021-07-13
Initial WM8960 Support

hftsai256:
Initial Waveshare WM8960 Audio HAT support. Requires a separated archive:
/lib/modules/5.4.83+/kernel/sound/soc/codecs/snd-soc-wm8960.ko 
/lib/modules/5.4.83+/kernel/sound/soc/bcm/snd-soc-wm8960-soundcard.ko
/boot/overlays/wm8960-soundcard.dtbo

Can those kernel modules be built under qemu-static environment?

### DIFF
--- a/app/plugins/system_controller/i2s_dacs/dacs.json
+++ b/app/plugins/system_controller/i2s_dacs/dacs.json
@@ -56,7 +56,8 @@
     {"id":"st400-dac-amp","name":"ST400 Dac (PCM5122) - Amp","overlay":"iqaudio-dacplus","alsanum":"2","mixer":"Digital","modules":"","script":"","needsreboot":"yes"},
     {"id":"taudac","name":"TauDAC - DM101","overlay":"taudac","alsanum":"2","mixer":"","modules":"","script":"","eeprom_name":"TauDAC-DM101","needsreboot":"yes"},
     {"id":"terraberry-dac2","name":"Terra-Berry DAC 2/3","overlay":"i-sabre-q2m","alsanum":"2","alsacard":"DAC","mixer":"","modules":"","script":"","needsreboot":"yes"},
-    {"id":"es90x8q2m-dac","name":"Volumio ESS 9028QM","overlay":"es90x8q2m-dac","alsanum":"2","mixer":"Digital","modules":"","script":"","i2c_address":"48","needsreboot":"no"}
+    {"id":"es90x8q2m-dac","name":"Volumio ESS 9028QM","overlay":"es90x8q2m-dac","alsanum":"2","mixer":"Digital","modules":"","script":"","i2c_address":"48","needsreboot":"no"},
+    {"id":"wm8960-soundcard","name":"Waveshare - WM8960","overlay":"wm8960-soundcard,i2s-mmap","alsanum":"","alsacard":"wm8960soundcard","mixer":"Playback","modules":"","script":"wm8960-unmute.sh","i2c_address":"1a","needsreboot":"yes"}
   ]},
   {"name":"Odroid C1+","data":[
     {"id":"odroid-hifi-shield","name":"HiFi Shield","overlay":"","alsanum":"2","mixer":"","modules":"","script":""}
--- a/app/i18n/strings_en.json
+++ b/app/i18n/strings_en.json
@@ -316,6 +316,8 @@
         "CHECKING_DUPLICATE_PLUGIN": "Checking for duplicate plugin",
         "CONFIRM_PLUGIN_INSTALL": "Install Plugin",
         "CONFIRM_PLUGIN_INSTALL_WARNING_MESSAGE": "Warning, community plugins are developed by third party developers and might result in unexpected behaviours and system instabilitiy. Are you sure you want to install this plugin?",
+        "PLUGIN_LOGIN_MESSAGE": "To access the plugins store you need to <a href=\"/myvolumio/signup\">login to MyVolumio</a>",
+        "PLUGIN_LOGIN": "Please login",
         "CONFIRM_PLUGIN_UNINSTALL": "Confirm plugin uninstall",
         "CONFIRM_PLUGIN_UNINSTALL_MESSAGE": "Are you sure you want to uninstall this Plugin?",
         "COPYING_PLUGIN_TO_LOCATION": "Copying plugin to location",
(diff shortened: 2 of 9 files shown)


app/plugins/system_controller/i2s_dacs/dacs.json
```diff
@@ -56,7 +56,8 @@
     {"id":"st400-dac-amp","name":"ST400 Dac (PCM5122) - Amp","overlay":"iqaudio-dacplus","alsanum":"2","mixer":"Digital","modules":"","script":"","needsreboot":"yes"},
     {"id":"taudac","name":"TauDAC - DM101","overlay":"taudac","alsanum":"2","mixer":"","modules":"","script":"","eeprom_name":"TauDAC-DM101","needsreboot":"yes"},
     {"id":"terraberry-dac2","name":"Terra-Berry DAC 2/3","overlay":"i-sabre-q2m","alsanum":"2","alsacard":"DAC","mixer":"","modules":"","script":"","needsreboot":"yes"},
-    {"id":"es90x8q2m-dac","name":"Volumio ESS 9028QM","overlay":"es90x8q2m-dac","alsanum":"2","mixer":"Digital","modules":"","script":"","i2c_address":"48","needsreboot":"no"}
+    {"id":"es90x8q2m-dac","name":"Volumio ESS 9028QM","overlay":"es90x8q2m-dac","alsanum":"2","mixer":"Digital","modules":"","script":"","i2c_address":"48","needsreboot":"no"},
+    {"id":"wm8960-soundcard","name":"Waveshare - WM8960","overlay":"wm8960-soundcard,i2s-mmap","alsanum":"1","alsacard":"wm8960soundcard","mixer":"Playback","modules":["i2c-dev","snd-soc-wm8960","snd-soc-wm8960-soundcard"],"script":"","i2c_address":"1a","needsreboot":"yes"}
```
hftsai256, line 60: I noticed `alsanum` is not actually used and it varies from SBC to SBC (1 on pi zero, 2 on pi 3B+/4).

ashthespy:
Small request for clarity don the line - could you move your changes to `i2s_dacs/index.js` into a new commit and force push? 

Out of curiosity, why does this DAC it require both modules and the overlay? The device tree should technically load the modules right?

hftsai256:
> Out of curiosity, why does this DAC it require both modules and the overlay? The device tree should technically load the modules right?

Well... Yes and no. I reached [this thread](https://github.com/respeaker/seeed-voicecard/issues/281#issuecomment-765763250) and checked the default kernel configuration. Turns out CONFIG_SND_SOC_WM8960 is not set by default and I believe the majority of Volumio users wouldn't bother (or don't know how) to build their own kernel for the device.

ashthespy:
> > Out of curiosity, why does this DAC it require both modules and the overlay? The device tree should technically load the modules right?
> 
> Well... Yes and no. I reached [this thread](https://github.com/respeaker/seeed-voicecard/issues/281#issuecomment-765763250) and checked the default kernel configuration. Turns out CONFIG_SND_SOC_WM8960 is not set by default 

Correct, but that is to compile the kernel module, which thanks to your other PR in the build repo will bake the required modules into the image.

The overlay normally should then be able to pick the compiled modules up, without having to additionally manually load them right?

hftsai256:
> The overlay normally should then be able to pick the compiled modules up, without having to additionally manually load them right?

Just confirmed simply loading the overlay will also pull in the related kernel module, which is `snd_soc_wm8960`. I guess the one I was missing `i2c-dev` so the device won't talk.

```
volumio@voluzero:~$ cat /etc/modules
i2c-dev
volumio@voluzero:~$ lsmod | grep wm8960
snd_soc_wm8960         40960  1
regmap_i2c             16384  1 snd_soc_wm8960
snd_soc_core          200704  4 snd_soc_simple_card_utils,snd_soc_bcm2835_i2s,snd_soc_wm8960,snd_soc_simple_card
snd_pcm                94208  5 snd_pcm_dmaengine,snd_soc_bcm2835_i2s,snd_bcm2835,snd_soc_core,snd_soc_wm8960
snd                    73728  12 snd_compress,snd_seq,snd_timer,snd_seq_device,snd_bcm2835,snd_soc_core,snd_soc_wm8960,snd_pcm
```
This kernel option is kinda likely being overlooked, because `CONFIG_SND_SOC_WM8960` option doesn't show up in any of `bcm{2709|2711|2835|rpi}_defconfig`. We could:

1. Host the kernel module separately as this PR is doing
2. Build our own flavored kernel
3. File an issue to raspberrypi/linux and request a fix upstream

ashthespy:
Great work, sorry if my comments have led you down the rabbit hole ;-)

Ideally, option 3 would be best, as `CONFIG_SND_SOC_WM8960 ` should also have a depend  that ensure `i2c-dev ` gets loaded.

Until it's fixed upstream, we could use the modules that you have built. 

I would really like to avoid custom kernels for the pi, will be less maintenance effort, we (well @gkkpch) already has multiple platforms that use special kernels.

hftsai256:
I poked around and was trying to figure out the "no sound" problem on my board while waiting reposes from the issue filed upstream. Turns out in addition to kernel modules (which the two PRs mentioned above have it implemented properly), [wm8960_asound.state from official repo](https://github.com/waveshare/WM8960-Audio-HAT/blob/master/wm8960_asound.state) has to be applied with a refresh (`alsactl nrestore`) in order to get the sound out. I'm a bit surprised about it because tuning those volumes in alsamixer won't help, and am curious about what's the magic under the hood to make it work.

In terms of implementation, we may download the file in Build/recipe/pi.sh and put it in place with a script defined in dacs.json. Is there a better way to do this?

hftsai256:
Finally this works OOB after initial setup wizard + reboot. We still have to host kernel artifacts individually as the fix is only applied to kernel 5.13.y at this moment. I've asked to apply it on kernel 5.4.y and still waiting for their reply.

ashthespy:
@hftsai256 Don't have hardware to test this -- but it looks good to me! 
See my comments over at https://github.com/volumio/Build/pull/513#issuecomment-889092133 we should be fine using the modules from [wm8960-rpi-modules](https://github.com/hftsai256/wm8960-rpi-modules) until we switch to 5.10.yy that has everything baked in.

PR #2129, opened 2021-07-28
Sync Buster/master to buster/dev

ashthespy:
This really should be the other way around..


volumio:
Yes, it should be the other way around... 

One question: what is on dev that we don't have on master?
Because if that's the case, we can just delete dev and do another branch from master

ashthespy:
>One question: what is on dev that we don't have on master?

Thankfully, looks like nothing ;-)
https://github.com/volumio/Volumio2/compare/buster/master...buster/dev

Will just create it fresh!

PR #2162, opened 2021-10-03
MPD plugin: Fix handling of URIs containing 'bbc'

patrickkfkan:
When parsing MPD's track info, the MPD plugin checks whether the stream URI contains the letters 'bbc'. If it does, it assumes the URI has got to be a BBC web radio stream. The plugin then proceeds to replace 'bbc' with 'BBC' in the `Name` property of the track info object. Now, while web radio streams will often have a `Name` tag, this is not always the case with other types of streams. With those streams that do not have a `Name` tag, parsing will fail resulting in Volumio not updating playback status nor proceeding to the next song after the current one finishes playing.

This PR fixes the issue by checking for the existence of the `Name` property in the track info object.


volumio:
Thanks!

